PR #17, opened 2024-09-17
add unsubscribe

maan2003:


examples/vite-core/src/wallet.ts
```diff
@@ -2,10 +2,8 @@ import { FedimintWallet } from '@fedimint/core-web'
...
 const wallet = new FedimintWallet()
 
-// try {
-wallet.open()
-// } catch (e) {
-//   console.warn('Failed to open wallet', e)
-// }
+function init(onOpen: (boolean)) {
+   kwallet.open().then(onOpen)
+}
```
alexlwn123, line 7: reverting this change due to some errors... will readd in follow-up

```suggestion
wallet.open()
```

alexlwn123 (approved):

PR #61, opened 2024-10-02
feat: add twoslash syntax highlight to docs

alexlwn123:
* Adds a vitepress plugin to enable rich code hints in docs site


alexlwn123:
TODO: fix version collision before merging

PR #66, opened 2024-10-04
chore: declare worker message types

okjodom:
Quickly found it difficult to keep track of the worker messages, and realized the messages will only grow as components like the wallet are built out. This PR introduces `enum WorkerMessageType` for the purposes of tracking and strong typing messages


alexlwn123 (approved):
Great contribution! So happy to see you here, jodom :)

PR #73, opened 2024-10-10
feat: tweaked invoice rpcs

Kodylow:
Adds support for tweaked invoice rpcs https://github.com/fedimint/fedimint/pull/6149


packages/core-web/src/services/LightningService.ts
```diff
@@ -47,6 +47,66 @@ export class LightningService {
...
+    expiryTime: number | null = null, // in seconds
+    publicKey: string,
+    index: number,
+    extraMeta: JSONObject = {},
+  ): Promise<CreateBolt11Response> {
+    await this.updateGatewayCache()
+    const gateway = await this._getDefaultGatewayInfo()
+    return await this.client.rpcSingle(
+      'ln',
+      'create_bolt11_invoice_for_user_tweaked',
```
alexlwn123, line 84: nit: `create_bolt11_incoice_tweaked`

packages/core-web/package.json
```diff
@@ -26,12 +26,14 @@
   "main": "./dist/index.js",
   "types": "./dist/index.d.ts",
   "dependencies": {
-    "@fedimint/fedimint-client-wasm-bundler": "workspace:*"
+    "@fedimint/fedimint-client-wasm-bundler": "workspace:*",
+    "secp256k1": "^5.0.0"
```
alexlwn123, line 30: it looks the crypto stuff is only being used in tests, so this dep should be moved to `devDependencies`

packages/core-web/src/utils/crypto.ts
```diff
@@ -0,0 +1,44 @@
+import * as secp256k1 from 'secp256k1'
```
alexlwn123, line 1: import the functions you're using, instead of `import *`... this prevents tree shaking from removing the unused functions

packages/core-web/src/utils/crypto.ts
alexlwn123, line 1: If this is only being used for tests, then it should be moved to the `src/test` directory

alexlwn123 (changes requested):
See comments

pnpm-lock.yaml
alexlwn123, line 1: prefer bumping deps in its own pr

packages/core-web/src/services/LightningService.ts
```diff
@@ -47,6 +47,66 @@ export class LightningService {
...
+        expiry_time: expiryTime,
+        user_key: publicKey,
+        index,
+        extra_meta: extraMeta,
+        gateway: gateway.info,
+      },
+    )
+  }
+
+  // Returns the operation ids of payments received to the tweaks of the user secret key
```
alexlwn123, line 97: What's being scanned exactly? (invoices, paid invoices, unpaid invoices, something else?)

packages/core-web/src/services/LightningService.ts
```diff
@@ -47,6 +47,66 @@ export class LightningService {
...
+  async scanReceivesForTweaks(
+    userKey: string,
+    indices: number[],
```
alexlwn123, line 100: What are the indicies?

packages/core-web/src/services/LightningService.ts
```diff
@@ -47,6 +47,66 @@ export class LightningService {
...
+        index,
+        extra_meta: extraMeta,
+        gateway: gateway.info,
+      },
+    )
+  }
+
+  // Returns the operation ids of payments received to the tweaks of the user secret key
+  async scanReceivesForTweaks(
+    userKey: string,
```
alexlwn123, line 99: is this just a pubkey? 

It doesn't look like the user provides a key in the `createInvoiceTweaked` function, so how does the user get this key?

PR #95, opened 2024-10-31
wip: feat: fedimint client rpc ng

maan2003:
- abstract away web worker transport
- ~~(wip) handle multiple federations~~ (delayed for some other time)


packages/core-web/src/rpc.ts
```diff
@@ -0,0 +1,207 @@
...
+
+// Handles communication with the wasm worker
+// TODO: Move rpc stream management to a separate "SubscriptionManager" class
+export class RpcClient {
+  private transport?: RpcTransport
+  private transportInit: RpcTransportInit
+  private requestCounter = 0
+  private requestCallbacks = new Map<number, (response: RpcResponse) => void>()
+  private initPromise?: Promise<void>
+  private clientName: string | undefined
```
maan2003, line 31: try to remove client name from here and pass it as argument to method where this is needed

maan2003: looks like only rpcStream and rpcSingle need it.

alexlwn123:
replaced by https://github.com/fedimint/fedimint-web-sdk/pull/159 and https://github.com/fedimint/fedimint-web-sdk/pull/158

PR #111, opened 2025-01-30
chore: fixes package.json breaking react example

wfzyx:
The react example wasn't building locally and at https://web.fedimint.org/examples/vite-react.html via StackBlitz


alexlwn123 (approved):
Nice catch! Thanks for the contribution!

wfzyx:
I missed the fact that the pipelines needs an up-to-date lock file on yaml format; that has been provided now;

alexlwn123 (approved):

Issue #129, opened 2025-03-27
Add Remix Example App


sethdivyansh:
We have a couple of examples for frameworks, but it'd be great to add one for Remix.

- [ ]  Add a Remix Example App
- [ ]  Add docs page containing the example
- [ ]  Update Getting Started instructions with Remix setup info

PR #130, opened 2025-03-27
feat: add remix example

sethdivyansh:
Fix: #129 

- Added remix example
- Documented the remix example giving an overview of remix application and running it.
- Modified the root package.json to run remix-app from root directory.


alexlwn123 (changes requested):
Thanks for the contribution! Apologies for the delay on review. 

I'm struggling to get the remix example working in my local dev. Seems like some errors relating to remix itself.

image

PR #131, opened 2025-03-27
chore: use normal nix shell, no rust toolchains needed

dpc:


flake.nix
```diff
@@ -13,26 +13,26 @@
...
       {
-        devShells = fmLib.devShells // {
-          default = fmLib.devShells.default.overrideAttrs (prev: {
+        devShells = {
+          default = pkgs.mkShell {
             nativeBuildInputs = [
               fedimint.packages.${system}.devimint
               fedimint.packages.${system}.gateway-pkgs
               fedimint.packages.${system}.fedimint-pkgs
+              pkgs.bitcoin
```
dpc, line 24: https://github.com/fedimint/fedimint/blob/8ccf1aff159e855c62b429dc21271f72935bbb14/nix/flakebox.nix#L235 will need to add back some of the stuff the fedimint dev shell was adding.

PR #138, opened 2025-05-13
chore: bump test timeout

elsirion:
We were seeing timeout failures https://github.com/fedimint/fedimint-web-sdk/actions/runs/14977939859/job/42074892204?pr=128


alexlwn123:
Current CI failures aren't being caused by timeouts... do we still want to do this?

https://github.com/fedimint/fedimint-web-sdk/actions/runs/15001845446/job/42166175687?pr=138#step:4:604

PR #155, opened 2025-05-22
chore: Bumped fedimint client wasm 

alexlwn123:
Bumped fedimint client wasm to [fedimint commit a0c64e374df5b061d1f6988ff720e2119013aa48](https://github.com/fedimint/fedimint/commit/a0c64e374df5b061d1f6988ff720e2119013aa48)


alexlwn123:
Size increase still isn't great... but much better than `12MB`

ref: https://github.com/fedimint/fedimint/pull/7479

@Harshdev098 for future reference, use `nix build '.#wasmBundle'` inside of `fedimint/fedimint` to build the wasm.

PR #163, opened 2025-06-09
Alex/compile wasm in ci

alexlwn123:
Moves wasm building to CI & removes the wasm-pack build artifacts from the repo.

This will require some tweaking to get right.

#### TODOs
- [x] Update flake to allow building wasm from this repo
- [x] Ensure both bundler and web targets are included (ref https://github.com/fedimint/fedimint/pull/7479)
- [x] Update release flow to build WASM before releasing packages
  - [x] Verify the Wasm packages get built/published properly
  - [x] Verify `core-web` gets built/published properly
- [x] Add caching to CI
- [x] Update size job to build wasm before measuring size
- [ ] Add local dev option to pull compiled package from CDN?


alexlwn123:
cc: @dpc

dpc:
I restarted that step to see how it works when ran second time. Obervations:

* It's weird that building is in the "Test" workflow.

* This job downloads some very chunky stuff, chrome browser, some automation stuff, the Fedimint binaries. The graphical programs pull in Gtk and graphical toolkit libraries, etc. Given the bandwidth limits between GHA datacenter and NixOs cache and Cachix, it is taking a while. (5 minutes and still going, as I'm writing it).

dpc:
image


Whatever this CI run builds is not being actually pushed to the cachix because PR builds have only read-only access. But once landed, this should change, so should be OK. But it's only like 1minute, relative to 14minutes it took to download all of it, so not very important.


Moving CI to run on self-hosted runners would eliminate downloading part and make it all run very snappily.

dpc:
image

This repo does already have access to self hosted runners.

PR #172, opened 2025-07-28
feat: Backup, Recovery, Transport Abstraction and Tauri initialisation

MrImmortal09:
Backup and recovery with example.

Currently work in progress


packages/core-web/src/types/rpc.ts
```diff
@@ -51,6 +52,10 @@ export type RpcRequest =
   | {
       type: 'get_mnemonic'
     }
+  | {
+      type: 'backup_to_federation'
+      metadata: JSONValue
+    }
```
maan2003, line 58: not needed? now that it is through client_rpc

MrImmortal09:
dev/recovery branch is the one with recovery/backup  working but it is not tested and can break. 
Closing this PR as a new architecture is arriving soon.
Snapshot for the recovery version:
`
npm i @fedimint/core-web@0.0.0-dev-recovery-f92131bde4b3ae5ef951de8668d8b304b4ab736a
`

**Please do not use this in the production or with real cash.**

PR #176, opened 2025-08-24
fix: Update listTransaction method

Harshdev098:
Have removed blind casting and improved type inferences


Harshdev098:
@alexlwn123 I have updated the PR

PR #194, opened 2025-10-01
Bump Wasm to support fedimint-client-rpc

MrImmortal09:


packages/types/src/rpc.ts
```diff
@@ -0,0 +1,111 @@
+import type { JSONValue } from './common'
```
MrImmortal09, line 1: will use this in future.

packages/core/src/FedimintWallet.ts
```diff
@@ -8,7 +8,7 @@ import {
   WalletService,
 } from './services'
 
-const DEFAULT_CLIENT_NAME = 'fm-default' as const
+const DEFAULT_CLIENT_NAME = 'dd5135b2-c228-41b7-a4f9-3b6e7afe3088' as const
```
alexlwn123, line 11: ???

packages/core/src/transport/TransportClient.ts
```diff
@@ -58,6 +61,48 @@ export class TransportClient {
     if (type === 'log') {
       this.handleLogMessage(message)
     }
+
+    if (type === 'rpcResponse') {
+      // Handle native RPC response format
+      const rpcResponse = (message as any).response
```
alexlwn123, line 67: unsafe. This will throw in runtime if message is undefined. we shouldn't need to cast to any.

packages/core/src/transport/TransportClient.ts
```diff
@@ -58,6 +61,48 @@ export class TransportClient {
...
+          } else if (rpcResponse.data !== undefined) {
+            callback({ data: rpcResponse.data })
+          } else {
+            // If it's a successful response without explicit data/error structure
+            callback({ data: rpcResponse })
+          }
+        }
+      }
+      return
+    }
```
alexlwn123, line 104: this code can be cleaned up. I'm a bit confused why it's all needed. We should have a standard structure for rpcResponse, right? So why is there handling for multiple response structures.

packages/core/src/transport/TransportClient.ts
```diff
@@ -188,14 +221,9 @@ export class TransportClient {
...
-  private async _rpcStreamInner<
-    Response extends JSONValue = JSONValue,
-    Body extends JSONValue = JSONValue,
-  >(
+  private async _rpcStreamInner<Response extends JSONValue = JSONValue>(
     requestId: number,
-    module: ModuleKind,
-    method: string,
-    body: Body,
+    rpcRequest: any,
```
alexlwn123, line 226: wrong... this shouldn't be any

packages/core/src/transport/TransportClient.ts
```diff
@@ -212,16 +240,23 @@ export class TransportClient {
         onEnd()
       }
     })
+
+    // Send the unified RPC request with specific type
     this.transport.postMessage({
-      type: 'rpc',
-      payload: { module, method, body },
+      type: rpcRequest.type as TransportMessageType,
```
alexlwn123, line 246: casting to the transport type? what's going on?

packages/core/src/transport/TransportClient.ts
```diff
@@ -232,13 +267,148 @@ export class TransportClient {
...
+  async sendUnifiedRpcRequest<T = JSONValue>(kind: any): Promise<T> {
+    const requestId = ++this.requestCounter
+
+    // Debug: Log what we're sending
+    this.logger.debug(
+      'TransportClient - sendUnifiedRpcRequest',
+      requestId,
+      kind,
+    )
+
```
alexlwn123, line 293: slop

packages/core/src/transport/TransportClient.ts
```diff
@@ -232,13 +267,148 @@ export class TransportClient {
...
+  }
+
+  // New RPC methods for the updated architecture - using existing transport infrastructure
+
+  /**
+   * Send a direct RPC request
+   */
+  async sendUnifiedRpcRequest<T = JSONValue>(kind: any): Promise<T> {
+    const requestId = ++this.requestCounter
+
```
alexlwn123, line 286: this function feels like AI slop. what're you trying to accomplish? What's the point of sendUnifiedRpcRequest? it doens't make sense to me at all.

PR #216, opened 2025-10-21
feat: add react native uniffi code gen

tobynwuba:


MrImmortal09:
Completed through https://github.com/fedimint/fedimint-sdk/pull/239

PR #230, opened 2025-11-25
feat: add parseOobNotes to walletDirector

### LINKED ISSUE #224: Move `parseNotes` from MintService to Core
If you want to build a "join federation when scanning ecash from unknown mint" UX you need to be able to decode the note without a wallet.

MrImmortal09:
Closes #224


elsirion (approved):
Let's gooooo! Do we need to bump any WASM reference @MrImmortal09?

PR #243, opened 2026-02-14
feat: implement onboarding flow with mnemonic generation and wallet integration

### LINKED ISSUE #237: Implement proper onboarding flow in web examples
### 1. Initial Check

**Trigger:** On Application Load

The application should check if the user has a mnemonic set (indicating an existing wallet).

* **Method:** Use `await director.hasMnemonicSet()`.

---

### 2. Onboarding Screen Logic

Based on the result of `hasMnemonicSet()`, the app will direct the user to one of two flows:

#### A. If `hasMnemonicSet()` returns `false`

**Action:** Display a clean, dedicated Onboarding screen where the user will set the mnemonic or generate it.

#### B. If `hasMnemonicSet()` returns `true`

**Action:** Skip the onboarding screen entirely.

---

### 3. General

* **Update Examples:** Ensure all project examples are refactored to utilize the updated functions and exports available in the new `packages/core`.

---

### 4. Goals

The main goal of this proposal is to create a seamless and intuitive user experience.

If you have ideas for further improvements, please share them.

Robinaditya1045:
closes #237

### DIFF
--- a/examples/next-js/app/globals.css
+++ b/examples/next-js/app/globals.css
@@ -172,3 +172,49 @@ button[type='submit'] {
   padding: 0.5rem;
   border-radius: 4px;
 }
+
+.onboarding-screen {
+  max-width: 600px;
+  margin: 2rem auto;
+  padding: 1rem;
+}
+
+.onboarding-card {
+  gap: 1rem;
+}
+
+.onboarding-actions {
+  display: flex;
+  flex-wrap: wrap;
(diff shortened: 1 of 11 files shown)


MrImmortal09 (changes requested):

MrImmortal09 (commented):
Image

The join federation creates a new client, you can't open a client which doesn't exist at all.

PR #252, opened 2026-03-10
docs: add versioning to docs

### LINKED ISSUE #168: Add Versioning to Docs
Since we're going through our first major breaking change, it'd be nice to keep a record of our old docs for prosperity.

See the Vitepress versioning plugin: https://vvp.imb11.dev/

Sky-walkerX:
It adds the versioning feature to the docs using vitepress-versioning-plugin as well as took the snapshot of current docs into 0.1.x directory, also converts the sidebar for multi version support accordingly.
Closes #168


alexlwn123 (commented):
Nice. Since there's only one version, it just redirects to the base path without a version. Can you add another (legacy) version, even just for testing, to show what it looks like to navigate to the old docs?

Also, I don't love how the vitepress version plugins requires us to copy over every file from the doc site for every new version... I'm wondering if we could have a section of live/evergreen docs and just just track any "old"/"legacy" docs between versions. Let me know if you have any ideas.

PR #262, opened 2026-03-27
feat(examples): add Expo Router wallet example with multi-screen fedimint flow

Sky-walkerX:
This PR adds a new Expo wallet example app to demonstrate production-style integration of the Fedimint SDK in a modern Expo setup.
It introduces a multi-screen architecture using Expo Router, shared wallet state helpers, and end-to-end flows for federation join, balance, history, send, receive, and wallet utilities.
Closes #258


MrImmortal09 (changes requested):
There is no onboarding workflow for this. Could you please create one?

You can use hasMnemonic to check whether a mnemonic is set or not, and then decide whether to show the onboarding flow accordingly.

Also, adding a 1–2 second splash screen when the app opens would be nice.

MrImmortal09:
The transaction history seems to have a bug. 

The transaction values are not shown correctly.

MrImmortal09 (changes requested):
Image

Operation failed while clicking the continue button.

Please do check the app flow by clearing app data before pushing.

MrImmortal09:
The balance isn't updating as well on payment of lightning invoice.

Sky-walkerX:
@MrImmortal09 I have tested the app, please review.

examples/expo-wallet/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "@fedimint/react-native": "workspace:*",
+    "expo": "~55.0.0",
+    "expo-clipboard": "~55.0.0",
+    "expo-constants": "~55.0.0",
+    "expo-linking": "~55.0.0",
+    "expo-router": "~55.0.0",
+    "expo-status-bar": "~55.0.0",
+    "react": ">=19.2.4",
+    "react-native": "0.83.1",
+    "react-native-fs": "^2.20.0",
```
MrImmortal09: Lets use expo-file-system instead of react-native-fs, The docs are depended on the expo version you are using so be careful.

MrImmortal09 (changes requested):
Would be better to rename it `expo-app` rather than `expo-wallet`.

examples/expo-wallet/package.json
MrImmortal09: Also make sure to update the Readme.md as well, do refer [this](https://github.com/fedimint/fedimint-sdk/blob/main/examples/react-native/README.md)

package.json
```diff
@@ -76,5 +76,11 @@
   },
   "simple-git-hooks": {
     "pre-commit": "pnpm lint"
+  },
+  "version": "1.0.0",
+  "dependencies": {
+    "expo": "~55.0.9",
+    "react": "19.2.0",
+    "react-native": "0.83.4"
```
MrImmortal09: Shouldn't be present in the root package.json.

examples/expo-app/README.md
```diff
@@ -0,0 +1,69 @@
...
+git submodule update --init --recursive
+```
+
+Then, ensure all dependencies are installed and the native bindings are built:
+
+```sh
+pnpm install
+nix develop .#android -c pnpm ubrn:android
+nix develop .#ios -c pnpm ubrn:ios
+pnpm build
```
MrImmortal09, line 30: add pnpm build:reactnative

Add this in the examples/react-native as well.

MrImmortal09 (changes requested):
A last few changes and we are good to go.

Nice PR Overall!

Sky-walkerX:
@MrImmortal09, made the changes.

Sky-walkerX:
Also I noticed that there are no docs for react-native and expo example app at [sdk-docs](sdk.fedimint.org). Should we create them?
If you prefer, I can do this in a follow-up docs-only PR.

MrImmortal09:
> Also I noticed that there are no docs for react-native and expo example app at [sdk-docs](sdk.fedimint.org). Should we create them? If you prefer, I can do this in a follow-up docs-only PR.

You can open an issue for this. I’ve am on the way to dev docs, so there’s no need to work on those. However, the regular/user-facing docs are something you can take up.

Sky-walkerX:
By user-facing docs, do you mean example section of sdk-docs?
image

MrImmortal09:
By user-facing i mean the docs required for people using the @fedimint/react-native from npm itself.

by dev docs I mean the docs required to build the bindings manually and work on them in this repo.

As of now I have a unification PR left, which will change the web interface, lets get that merged first. Then you can start working on the docs.

MrImmortal09:
@Sky-walkerX The app is crashing for me after the last commit, please have a look at it once.

Sky-walkerX:
> @Sky-walkerX The app is crashing for me after the last commit, please have a look at it once.

Sure, let me see.

Sky-walkerX:
@MrImmortal09 It's working fine now, it was `expo-file-system` version issue.
For now I have used expo-file-system/legacy to confirm it was the only root cause.

Should we move to the modern expo-file-system API, or stick with this for now?

examples/expo-app/src/wallet.ts
```diff
@@ -1,8 +1,10 @@
 import WalletDirector from '@fedimint/react-native'
 import type { FedimintWallet } from '@fedimint/react-native'
-import * as FileSystem from 'expo-file-system'
+import * as FileSystem from 'expo-file-system/legacy'
```
MrImmortal09: Don't import legacy, use simple expo-file-system, The docs are bit tricky so refer by the expo version you are using.

MrImmortal09 (changes requested):

examples/expo-app/src/wallet.ts
Sky-walkerX: I guess the syntax is same just had to change the import line. I have tested the app again, its working fine.

MrImmortal09 (approved):
LGTM.

Great Work @Sky-walkerX

PR #275, opened 2026-04-15
fix: correct RPC stream service for subscribeDeposit method in Wallet…

### LINKED ISSUE #274: subscribeDeposit should use the `wallet` module instead of `ln`
`WalletService.subscribeDeposit()` in `fedimint-sdk` currently calls `rpcStream('ln', 'subscribe_deposit', ...)`.

For onchain deposits this should call the `wallet` module instead:

```ts
return this.client.rpcStream(
  'wallet',
  'subscribe_deposit',
  { operation_id },
  this.clientName,
  onSuccess,
  onError,
)
```

## Current code

File:
`packages/core/src/services/WalletService.ts`

Current implementation:

```ts
return this.client.rpcStream(
  'ln',
  'subscribe_deposit',
  { operation_id: operation_id },
  this.clientName,
  onSuccess,
  onError,
)
```

## Why this is a bug

`subscribeDeposit()` belongs to `WalletService` and is intended for onchain deposit updates. Routing this RPC through the Lightning module is incorrect.

## Expected behavior

`WalletService.subscribeDeposit()` should stream deposit updates from the `wallet` module.

ngutech21:
fixes #274

### DIFF
--- a/.changeset/young-geese-retire.md
+++ b/.changeset/young-geese-retire.md
@@ -1,0 +1,5 @@
+---
+'@fedimint/core': patch
+---
+
+Fix WalletService.subscribeDeposit to call the wallet module instead of ln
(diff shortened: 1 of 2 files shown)


MrImmortal09 (approved):
LGTM

PR #279, opened 2026-04-20
feat: add getMetaConsensusValue method and documentation

ngutech21:
This adds a small SDK wrapper for reading federation metadata from the meta module through the existing client RPC path. The change exposes federation.getMetaConsensusValue(), exports DEFAULT_META_KEY, wires meta into the typed module list, and adds coverage for the new RPC call plus docs.

This PR only adds read access for get_consensus_value; it does not introduce a separate MetaService or any of the other meta RPCs.

Note: this still needs a WASM bump after [fedimint/fedimint#8532](https://github.com/fedimint/fedimint/pull/8532) is merged, since the SDK side depends on that RPC being available in the bundled WASM.


.changeset/better-banks-pay.md
```diff
@@ -0,0 +1,5 @@
+---
+'@fedimint/core': minor
```
MrImmortal09, line 2: The wasm would also require a version bump.

MrImmortal09: as we have updated the ref in flake.nix

MrImmortal09 (changes requested):

.changeset/better-banks-pay.md
ngutech21: good catch. added the wasm bump to the changeset

MrImmortal09 (approved):
LGTM

PR #283, opened 2026-04-30
feat(rn-bindings): build via Nix instead of in-shell cargo cross-compile 

dpc:
Adds a 'ubrn:nix:<platform>:release' npm script that:

  1. Pre-builds each Rust target as a Nix derivation
     (#android-<triple> / #ios-<triple>) and copies the resulting
     artifacts into the cargo target directory.
  2. Runs 'ubrn build <platform> --no-cargo --and-generate --release'
     so ubrn skips its own cargo cross-compile entirely and only
     does the JNI-libs / xcframework assembly + TS/cpp/Kotlin binding
     generation.

The original 'ubrn:android' / 'ubrn:ios' scripts are kept untouched
for use cases that need the in-shell flow (e.g. local edit-test loops
that benefit from cargo's incremental rebuilds inside the source tree).

Net effect: CI runners no longer rebuild rocksdb, aws-lc-sys, tokio,
or the fedimint-* crates; they fetch them from cachix in seconds.


MrImmortal09 (approved):
Appreciate this a lot, learned quite a bit from this implementation

PR #293, opened 2026-05-18
feat: add forceRecover option to joinFederation

ngutech21:
Adds a `forceRecover` option to `joinFederation`, while keeping the existing `joinFederation(inviteCode, clientName)` call working.

This is needed when restoring a wallet: after joining the federation again, `forceRecover` lets the client trigger recovery so the wallet balance can be restored.

### DIFF
--- a/.changeset/join-federation-force-recover.md
+++ b/.changeset/join-federation-force-recover.md
@@ -1,5 +1,5 @@
 ---
-'@fedimint/core': minor
+'@fedimint/core': patch
 ---
 
 Added `forceRecover` support to `FedimintWallet.joinFederation()` through a backwards-compatible options object.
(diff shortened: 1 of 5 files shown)


.changeset/join-federation-force-recover.md
```diff
@@ -0,0 +1,5 @@
+---
+'@fedimint/core': minor
```
MrImmortal09: Should be patch.

MrImmortal09 (changes requested):

.changeset/join-federation-force-recover.md
ngutech21: fixed

MrImmortal09 (approved):
LGTM.

PR #327, opened 2026-08-06
Fix: Map swallowed SDK string errors to user friendly UI messages in Vite demo (`MnemonicManager`)

### LINKED ISSUE #326: Vite example app: All SDK validation errors are swallowed as Operation failed
### Problem Summary
In the `vite-core` example app, when the SDK rejects an action (like generating or setting a mnemonic), the UI completely swallows the actual error reason and displays a generic `Operation failed` message to the user.

For example, all of the following errors are successfully logged in the console, but the user only sees `"Operation failed"`:
- `Error setting mnemonic: the mnemonic has an invalid checksum`
- `Error generating mnemonic: Wallet mnemonic already exists. Please clear existing data before generating a new mnemonic.`
- `Error setting mnemonic: invalid word count`

### Cause
When the WASM SDK throws an error, it throws a raw primitive string. However, the `extractErrorMessage(error)` helper function in `examples/vite-core/src/App.tsx` expects errors to strictly be `Error` instances or RPC objects. 
Because `typeof error === 'string'` is completely unhandled, the switch statement falls through to the default fallback: `"Operation failed"`.

Ayush4958:
### **Summary**
Fixes an issue in the `vite-core` example app where raw SDK string errors were being swallowed and surfaced to the user as generic `"Operation failed"` messages. 

**Closes #326**

### Implementation details
- Updated the `extractErrorMessage` helper inside the `MnemonicManager` component to properly intercept primitive string errors thrown by the WASM modules.
- Added a `mapTechnicalErrorToFriendly` dictionary to translate highly technical backend strings into clean, user friendly UI messages.

### DIFF
--- a/js/examples/vite-core/src/App.tsx
+++ b/js/examples/vite-core/src/App.tsx
@@ -107,41 +107,21 @@ const MnemonicManager = () => {
 
   const clearMessage = () => setMessage(undefined)
 
-  const mapTechnicalErrorToFriendly = (rawError: string): string => {
-    const lowerError = rawError.toLowerCase()
-    
-    if (lowerError.includes('mnemonic already exists')) {
-      return 'You already have a wallet set up.'
-    }
-    if (lowerError.includes('invalid checksum')) {
-      return 'The recovery phrase you entered is invalid. Please double-check your spelling.'
-    }
-    if (lowerError.includes('invalid word count')) {
-      return 'Recovery phrases must be exactly 12 or 24 words long.'
-    }
-    if (lowerError.includes('contains an unknown word')) {
-      return 'The recovery phrase contains an unknown word. Please check your spelling.'
-    }
-    
-    // Fallback for unknown strings
-    return 'Operation failed'
-  }
-
-  // Helper function to extract user-friendly error messages
+  // Helper function to extract error messages directly
   const extractErrorMessage = (error: any): string => {
     let errorMsg = 'Operation failed'
 
     if (typeof error === 'string') {
-      errorMsg = mapTechnicalErrorToFriendly(error)
+      errorMsg = error
     } else if (error instanceof Error) {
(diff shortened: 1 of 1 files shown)


examples/vite-core/src/App.tsx
```diff
@@ -107,19 +107,41 @@ const MnemonicManager = () => {
...
+      return 'Recovery phrases must be exactly 12 or 24 words long.'
+    }
+    if (lowerError.includes('contains an unknown word')) {
+      return 'The recovery phrase contains an unknown word. Please check your spelling.'
+    }
+    
+    // Fallback for unknown strings
+    return 'Operation failed'
+  }
+
```
MrImmortal09: @Ayush4958 Thanks for the PR. 
I think we should remove this as if the ui msgs are to more friendly then i think it should be done in fedimint/fedimint, rather than doing it here.

MrImmortal09 (changes requested):

examples/vite-core/src/App.tsx
MrImmortal09: This way it would be made sure that the change is uniform across every example we have.

MrImmortal09: Also I feel that even the shortend format of msgs is sufficient.

Ayush4958: > @Ayush4958 Thanks for the PR. I think we should remove this as if the ui msgs are to more friendly then i think it should be done in fedimint/fedimint, rather than doing it here.

Gotcha should I raise PR in fedimint/fedimint ??
should I close this PR ?? and open one in fedimint for it

MrImmortal09: For now you can just make sure that the error is shown to the user instead of just operation failed.

MrImmortal09 (approved):
LGTM

MrImmortal09 (approved):
LGTM

PR #331, opened 2026-08-12
docs: clarify WalletDirector and wallet lifecycle

ngutech21:
- Updated the docs to present `WalletDirector.createWallet()` as the supported public API, replacing direct `FedimintWallet` construction.
- Clarified that `wallet.cleanup()` terminates the director-owned transport and invalidates all associated wallets.
- Documented that lazy directors using a custom `dbPath` must call `initialize(dbPath)` before `createWallet()`.
- Added corresponding wallet creation and lifecycle examples.

No runtime behavior was changed. Type checking, formatting, and the documentation build pass.


MrImmortal09 (approved):
LGTM

PR #332, opened 2026-08-12
fix: remove obsolete vitest hack

ngutech21:
PR #6569 landed a long time ago and has been included in Vitest since v2.1.2. Since we use Vitest v3.2.4, this workaround is obsolete and can be safely removed.


MrImmortal09 (approved):
LGTM

PR #333, opened 2026-08-12
chore: remove obsolete Bun dependency and lockfile

ngutech21:
`@types/bun` supported a now-deleted `Bun.file(...)` script and became unused after commit `cf18ee9`. The repository exclusively uses pnpm, while the unmaintained `bun.lockb` from commit `9d2d350` is not used by any workflow. Removing both does not affect Bun support for consumers because `create-fedimint-app` only generates Bun-compatible commands.

Fix:
- Remove the unused `@types/bun` development dependency.
- Remove its associated pnpm lockfile entries.
- Delete the obsolete `bun.lockb`.


MrImmortal09 (approved):
LGTM

PR #339, opened 2026-08-18
Move fedimint-sdk-ffi here

zeenix:
Since fedimint-sdk-ffi isn't used by anyone else directly than this SDK, it makes sense to just
have it here directly. This still doesn't stop anyone from using it in external projects, if they
want to, but that's doubtful given it's just a bridge.

Everything from that repo comes along: the `fedimint-client-uniffi` crate (with its README and
.gitignore) and the Rust CI and Nix build workflows, whose path filters and flake output names
happen to match this repo's layout as-is. The crate is imported at fedimint-sdk-ffi's current main
(`505852f`) with the still-open fedimint/fedimint-sdk-ffi#17 applied on top, so that PR and the
repo itself can be retired once this lands. Deliberately not imported as files: the ffi repo's
flake.nix/flake.lock — instead the final commit ports their build derivations into `nix/ffi.nix`,
building from the in-tree crate, and drops the external flake input entirely — and the codex
workflows (already present here).

Since neither side of the JS-to-Rust seam had any tests, two commits add coverage that runs on
every PR touching these paths:

- Rust smoke tests for the crate: handler construction over rocksdb in a temp dir, the
  invalid-JSON and unusable-db-path error paths, a mnemonic set/get round-trip through the
  database, and handler errors surfacing as `error` responses followed by the `end` marker. All
  offline-capable.
- Unit tests for `ReactNativeTransport` against a stubbed `RpcHandler` (injected via a Vitest
  alias, so no generated bindings or native library is needed). The wire-format test pins the JS
  half of the same RPC protocol contract the Rust tests pin from the native side.

Verified beyond CI: `cargo test`/clippy/fmt pass locally, the Vitest suite passes in under a
second after a plain `pnpm install`, and `cargo metadata` confirms ubrn's `--no-cargo` flow
resolves the crate's target directory exactly where `nix-prebuild.sh` places the Nix-built libs.

Generated by Claude Fable 5.


MrImmortal09 (approved):
LGTM

PR #345, opened 2026-08-20
Reduce build downloads and keep the FFI cache warm

zeenix:
Follow-up to #339, rebased on main now that it landed.

Building the RN bindings locally used to mean multi-GB downloads and a from-source cross-compile
of ~760 derivations, because nothing kept the Cachix cache warm and the dev shells carried weight
the builds never use. This PR attacks both:

Smaller closures:
- Drop Android emulator system images (multi-GB per ABI; nothing here runs an emulator) and the
  Playwright browser bundles from the android/iOS dev shells — the wasm shells keep their own.
- Drop the armv7/i686 Android targets nothing ships from both toolchains (verified the trim only
  removes those targets' env vars from the build derivations).
- Scope the Nix build source to files the build actually reads, and keep FFI evaluation lazy so
  wasm-only builds don't pay for it.

Warm cache:
- Explicitly `cachix push` the FFI lib outputs **and their crane deps derivations** (~95% of the
  build cost) after CI builds — cachix-action's post-run hook never re-pushes paths already
  realized on the warm self-hosted runner, which is how the cache went cold. Same treatment for
  `wasmBundle`, whose auth token was never actually reaching the reusable verify workflow.
- Pin fenix like the other FFI inputs so a routine lock refresh can't silently invalidate every
  toolchain and cross-compile derivation.

Review hardening (from an adversarial review pass over the branch):
- rust-ci: `--locked` so CI validates the resolution the Nix builds ship, dead CLASSPATH/feature
  config dropped, no redundant build step, lint job unserialized; both imported workflows get
  main-only push triggers, concurrency groups, timeouts, and the iOS job the same parallelism cap
  the Justfile uses against macos-latest OOM.
- The RN transport no longer logs request/response payloads (set_mnemonic/get_mnemonic carried
  wallet words into device logs), and its tests are now also type-checked in CI against the
  stubbed bindings.
- The FFI crate declares the tokio features it actually uses, reruns build.rs when the sdallocx
  stub changes, and degrades to an error response instead of aborting the host app if RPC
  response serialization ever fails.

Known deliberate skips: the in-flight-RPC cancellation on `cleanup` (inherited design, needs its
own discussion), the possibly-forcer aws-lc-rs/socket2 direct deps (removing them needs
cross-target build verification), and deduplicating the shipped-target lists across nix/yaml/sh
(fits the planned repo restructuring).

Generated by Claude Fable 5.

### DIFF
--- a/.github/workflows/nix-build.yml
+++ b/.github/workflows/nix-build.yml
@@ -2,6 +2,7 @@ name: Nix Native Builds
 
 on:
   push:
+    branches: [main]
     paths:
       - 'fedimint-client-uniffi/**'
       - 'nix/**'
@@ -16,10 +17,17 @@ on:
       - 'flake.lock'
       - '.github/workflows/nix-build.yml'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+permissions: {}
+
 jobs:
   build-ios:
     name: Build iOS Artifacts
     runs-on: macos-latest
+    timeout-minutes: 240
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
 
@@ -27,8 +35,13 @@ jobs:
         uses: cachix/install-nix-action@630ae543ea3a38a9a4166f03376c02c50f408342 # v31
         with:
           nix_path: nixpkgs=channel:nixos-unstable
+          # max-jobs/cores mirror the Justfile's build-ios: heavy native deps
+          # (rocksdb, aws-lc-sys) otherwise OOM macos-latest's small runners
+          # on cache-cold builds. Warm runs only substitute, so unaffected.
           extra_nix_config: |
             sandbox = relaxed
+            max-jobs = 1
+            cores = 1
 
       - name: Setup Cachix
         uses: cachix/cachix-action@18cf96c7c98e048e10a83abd92116114cd8504be # v14
@@ -51,6 +64,7 @@ jobs:
   build-android:
     name: Build Android Artifacts
     runs-on: ubuntu-latest
+    timeout-minutes: 90
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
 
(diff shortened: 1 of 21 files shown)


.github/workflows/nix-build.yml
```diff
@@ -16,19 +17,31 @@ on:
...
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+permissions: {}
+
 jobs:
   build-ios:
     name: Build iOS Artifacts
     runs-on: macos-latest
+    timeout-minutes: 240
```
MrImmortal09, line 32: Isn't 240 minutes too much? Let's reduce it to 120 minutes if that works for you. The last run took about 74 minutes. Additionally, this runs on GitHub-provided runners, which charge differently.

.github/workflows/rust-ci.yaml
```diff
@@ -7,18 +7,26 @@ name: Rust Continuous Integration
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
 
 permissions: {}
 
 jobs:
   test:
     name: 'Run tests'
     runs-on: ubuntu-24.04
```
MrImmortal09, line 28: lets run it on self hosted runners

.github/workflows/nix-build.yml
```diff
@@ -51,6 +64,7 @@ jobs:
   build-android:
     name: Build Android Artifacts
     runs-on: ubuntu-latest
```
MrImmortal09: can we run this too on self hosted runner?

.github/workflows/rust-ci.yaml
```diff
@@ -32,16 +40,13 @@ jobs:
...
-        run: cargo build
-
       - name: 'Test'
-        run: CLASSPATH=./tests/jna/jna-5.14.0.jar cargo test --features uniffi/bindgen-tests
+        run: cargo test --locked
 
   check:
     name: 'Run clippy and fmt'
-    needs: test
     runs-on: ubuntu-24.04
```
MrImmortal09, line 48: same

MrImmortal09 (changes requested):
Let’s shift to self-hosted runners wherever possible. We can’t do much about the iOS builds, which could still cost us a lot since they run on every commit pushed across PRs for UniFFI. While this might seem minor right now, it’s better to be cautious before costs increase exponentially.

MrImmortal09 (approved):
LGTM

PR #353, opened 2026-08-23
[WIP] React Native e2e tests 

MrImmortal09:


MrImmortal09:
@fedimint-bot create a PR description for this PR.

MrImmortal09:
@fedimint-bot review.

MrImmortal09:
@fedimint-bot review.

MrImmortal09:
@fedimint-bot review

MrImmortal09:
@fedimint-bot review

MrImmortal09:
@fedimint-bot review

scripts/e2e-android/setup-and-start-appium.sh
```diff
@@ -0,0 +1,111 @@
...
+    APPIUM_LOG_LEVEL="debug"
+  else
+    APPIUM_LOG_LEVEL="info"
+  fi
+
+  ATTEMPT_LOG="$APPIUM_HOME/appium-attempt-${attempt}.log"
+  nohup appium --port "$CANDIDATE_PORT" \
+    --log-level "$APPIUM_LOG_LEVEL" \
+    --allow-insecure=uiautomator2:adb_shell,uiautomator2:chromedriver_autodownload \
+    >"$ATTEMPT_LOG" 2>&1 </dev/null &
```
zeenix: **[P1] Bind the shell-enabled Appium server to loopback**

Appium's default listen address is `0.0.0.0`, and this invocation enables
`uiautomator2:adb_shell` without authentication. On a developer machine with a reachable
network interface, another machine can create a session and execute arbitrary ADB shell
commands on an attached device. The server also survives the test run: teardown deletes
the session, while this `nohup` process keeps running.

Please pass `--address 127.0.0.1`; both the health check and `AppiumManager` already connect
there, so no remote access is needed.

Sources: [server defaults][defaults] and [ADB shell access][security].

[defaults]: https://appium.io/docs/en/3.0/reference/cli/server/
[security]: https://appium.io/docs/en/3.0/guides/security/

*Posted by Codex on behalf of @zeenix.*

rust/fedimint-sdk/src/android.rs
```diff
@@ -0,0 +1,243 @@
...
+    // empty and would abort otherwise.
+    //
+    // `into_raw` deliberately leaks the global reference. `ndk_context` hands
+    // out copies of the pointer for the rest of the process's life and has no
+    // way to signal that the last reader is done, so there is no point at
+    // which deleting it would be safe. It is the *application* context, which
+    // lives as long as the process regardless, so the leak costs one
+    // reference rather than keeping an Activity alive.
+    unsafe {
+        ndk_context::initialize_android_context(vm.cast(), context.into_raw().cast());
```
zeenix: **[P2] Declare the permission required to read Android's DNS configuration**

Publishing the context is only half of this setup. The resolved `hickory-resolver 0.26.2`
calls `ConnectivityManager.getActiveNetwork()` and `getLinkProperties()`, which require
`android.permission.ACCESS_NETWORK_STATE`, but the SDK's manifest declares only `INTERNET`.

For an embedding app without that additional permission, the JNI call throws a
`SecurityException`; `iroh-dns 1.1.0` then falls back to Google's DNS servers. This means
the platform DNS path never works for that host, and iroh connectivity can fail on networks
where external DNS is blocked or the configured resolver is required. The current E2E
run can pass through the HTTP gateway without detecting this.

Please add `ACCESS_NETWORK_STATE` to the SDK's manifest so it is merged into consuming apps,
and verify that the resolver can actually read the device's DNS configuration.

[Android API reference][connectivity]

[connectivity]: https://developer.android.com/reference/android/net/ConnectivityManager

*Posted by Codex on behalf of @zeenix.*

rust/fedimint-sdk/src/android.rs
```diff
@@ -0,0 +1,243 @@
...
+    init_logging();
+
+    PUBLISHED.call_once(|| {
+        // Logged before the attempt, not only after it. Whether this function
+        // runs at all is a property of how the host loaded the library —
+        // `System.loadLibrary` calls `JNI_OnLoad`, a bare `dlopen` does not,
+        // and JNA reaches for `dlopen` first — so "did we get here" is worth
+        // being able to answer separately from "did it work".
+        info!("JNI_OnLoad: publishing the Android context");
+        match publish(vm) {
```
zeenix: **[P2] Allow context initialization to recover after an early library load**

If an `Application` field initializer or `attachBaseContext()` calls a synchronous binding
such as `InviteCode.parse`, this hook runs before `ActivityThread.mInitialApplication` is
assigned. Both lookups then return null: AOSP's `AppGlobals.getInitialApplication()` simply
calls `ActivityThread.currentApplication()`, so it is not an earlier-populated fallback.

`publish` returns an error, but this closure returns normally and permanently completes
`PUBLISHED`. Later SDK creation only calls `init_logging()`, and loading the same library
again does not rerun `JNI_OnLoad`. Consequently, an otherwise valid later join/receive
through iroh still aborts on the uninitialized `ndk_context`.

Please retain the VM and retry publication once an Application is available, marking it
initialized only after success (or provide an explicit context initialization path).
Cover loading a binding before Application attachment, then using the SDK after startup.

*Posted by Codex on behalf of @zeenix.*

zeenix (changes requested):
Reviewed `295873d663db8a77d8ac9624c4c6085e40f75786` with Codex on behalf of @zeenix.

Three findings inline: the Appium server exposes ADB shell access on all interfaces,
the new DNS path is missing its manifest permission, and an early library load permanently
disables context initialization.

Validation: TypeScript checking and shell syntax checks passed locally. The current Kotlin
SDK CI run passed, including all five Appium tests; those tests do not cover the cases above.

Please also split `90890e8` into separate logical commits for Android context/loader wiring,
logcat support, and the Appium scrolling fixes.

scripts/e2e-android/run-android-e2e.sh
```diff
@@ -0,0 +1,305 @@
...
+  echo "Booting AVD: $AVD_NAME"
+  nohup emulator -avd "$AVD_NAME" -no-snapshot -no-boot-anim -no-window -wipe-data -gpu swiftshader_indirect \
+    >"$LOG_DIR/emulator.log" 2>&1 &
+  local emu_pid=$!
+
+  # One loop for the transport and the boot both: `getprop` against an absent
+  # device simply fails, so waiting on the property covers `wait-for-device`
+  # too, and there is only one wait left to bound.
+  local deadline=$((SECONDS + BOOT_TIMEOUT_SECS))
+  until [[ "$(adb shell getprop sys.boot_completed 2>/dev/null | tr -d '\r')" == "1" ]]; do
```
zeenix: **[P2] Poll the newly launched emulator by serial**

The interactive menu allows booting an AVD while another phone or emulator is attached,
but this unqualified `adb shell` does not identify the emulator just started. It can read
`sys.boot_completed=1` from the existing device and return before the new emulator boots.
Once both devices are visible, it instead fails with "more than one device/emulator";
the error is discarded, so the loop eventually times out and kills the healthy emulator.

Please identify the new emulator's serial (for example, by assigning its console port)
and use `adb -s <serial>` for every boot probe. I reproduced the timeout using the actual
`boot_avd` function with mocked ADB responses for two ready devices.

*Posted by Codex on behalf of @zeenix.*

zeenix (commented):
One additional finding on `295873d`: the emulator boot check needs to select its device
explicitly. The other three findings are already covered by the existing review.

*Posted by Codex on behalf of @zeenix.*

PR #355, opened 2026-08-24
Capture wasm worker crash stacks and add a native repro for the peg_in crash

zeenix:
Follow-ups from tracking down the `generateAddress` flake (#330) to its actual root cause: **`WalletClientModule::supports_safe_deposit()` panics on a recoverable db `WriteConflict`** when `peg_in` races the wallet module's post-join background writes — filed upstream with a full backtrace as fedimint/fedimint#9046 (and the unwind-turned-abort in `ClientHandle::drop` as fedimint/fedimint#9053).

- **transport-web**: include the error's stack when reporting uncaught worker errors. `event.message` alone is how CI ended up with a bare `Uncaught RuntimeError: unreachable`; with the stack, the trap carries its wasm frames (this is what made the diagnosis possible). Patch changeset included.
- **fedimint-client-uniffi**: opt-in native reproduction test for the crash (set `REPRO_INVITE_CODE` and run under devimint; self-skips otherwise — CI unaffected). It reproduces the panic deterministically with a symbolized backtrace today, and becomes the verification that the upstream fix works once the pins are bumped.
- **integration-tests**: the 100ms-sleep workaround now documents the real root cause instead of the old investigate-me TODO.

Verified: all 12 unit/RN vitest tests and the 5 crate tests pass; prettier and cargo fmt clean.

### DIFF
--- a/.changeset/worker-error-stacks.md
+++ b/.changeset/worker-error-stacks.md
@@ -1,0 +1,8 @@
+---
+'@fedimint/transport-web': patch
+---
+
+Include the error's stack when reporting uncaught wasm worker errors.
+
+`event.message` alone drops the stack, and for a wasm trap the stack (with its wasm frame
+references) is the only clue to what crashed.
--- a/fedimint-client-uniffi/src/lib.rs
+++ b/fedimint-client-uniffi/src/lib.rs
@@ -237,6 +237,54 @@ mod tests {
         assert_eq!(responses[0]["data"], json!(true));
     }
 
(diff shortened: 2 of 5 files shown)


Ayush4958 (changes requested):
I think we should Refactor the `peg_in_right_after_join` test to use `#[ignore]` instead of a silent return on a missing env var. 

Also we should clean up the string concatenation in `worker.js` to prevent duplicate err msg while preserving the stack trace.

MrImmortal09 (approved):
LGTM

PR #314, opened 2026-07-25
ci: unify PR workflows with dynamic path filtering

Ayush4958:
### Description
I had optimizes the CI pipeline by unifying `pull-request.yml` &  `react-native-pr.yml` into a single, dynamic workflow, it intelligently detects which domains were modified & only runs the relevant jobs, significantly reducing CI run times & queue congestion.

### Changes Made
- **Consolidated Workflows** :- Deleted `react-native-pr.yml` and moved RN build jobs into `pull-request.yml`.
- **Dynamic Path Filtering** :- Introduced `detect-changes` to map file modifications to specific domains (`core`, `web`, `docs`, `cli`, `react-native`).
- **Conditional Job Execution** :- Updated `verify.yml` with boolean inputs (`run_core`, `run_web`, `run_docs`, `run_cli`) so heavy jobs like Playwright WASM tests or 2 hour iOS builds are skipped when only unrelated files (e.g. docs) change.
- **Added Missing Coverage** :- Added a `test-cli` job to run the `create-fedimint-app` vitest suite, which was previously missing from CI.
- **Fail safe Triggers** :- Changes to `.github/workflows/**` force all path flags to `true` to ensure workflow logic edits are thoroughly verified.

**Closes #311**

### Note on Skipped Jobs
Jobs skipped through conditionals will report as "Skipped" rather than "Success". If there are strict branch protection rules requiring specific check names, they may need to be set to "Required when available" instead of strictly "Required".

PR #373, opened 2026-09-04
ci: replace paths-filter with local change detection

### LINKED ISSUE #311: Prop: Optimize and unify PR workflows using dynamic path filtering
## Description

Currently, CI runs are split across two separate workflow files
- `pull-request.yml` (Web / Core) :- Triggered via `paths-ignore` (skips React Native).
- `react-native-pr.yml` (Mobile) :-Triggered via `paths` (only runs for React Native files).

This creates problem While splitting these workflows was a good step toward isolation, but the rigid `paths`/`paths-ignore` system at the GitHub actions trigger creates a few inefficienciesheckout, nix setup, and dependency installations independently.

## Proposed Solution
We can consolidate the PR CI into a single `pull-request.yml` utilizing a dynamic path detection step

**How it would work :-**
1. A lightweight `detect-changes` job runs first.
2. It outputs boolean flags mapped to domains (`core`, `web`, `react-native`, `docs`).
3. Each downstream job uses an `if` condition based on those flags.

**Example :-**
- If only `docs/**` change → run `lint` and `build-docs`. Skip tests and RN builds.
- If only `packages/react-native/**` change → run `lint` and RN Android/iOS builds. It will Skip the 30 min web/WASM tests.
- If `packages/core/**` change → run everything.
- If multiple domains change, it simply combines the requirements and runs the necessary jobs for all the changed domains, while safely skipping the unrelated ones.

dpc:
*Posted by Tau*

### Summary

Replace the third-party path-filter action with repository-owned Git diff detection while consolidating React Native verification into the main pull-request workflow. Previously, the reusable verification suite ran for every pull request and React Native work ran in a separate workflow; now the workflow selects the affected core, web, docs, CLI, and React Native jobs without giving an external action access to pull-request metadata. This keeps the existing checks available while removing that dependency.

### Details

The detector compares the pull request's merge-base range or a main-branch push range with a NUL-delimited `git diff --no-renames --name-only`. Shared workspace configuration and CI implementation changes enable every domain, while source paths enable the jobs that depend on them. Manual runs and unavailable revision ranges fail closed by running every domain, and treating renames as a deletion plus an addition keeps both affected domains covered. The consolidated workflow retains main-branch React Native safety-net builds, React Native transport checks, and the `ci-ios` PR label gate; label-only runs use isolated concurrency so unrelated labels do not cancel an in-flight verification run.

This is an alternative to #314 and closes #311.

### Reviews

An independent CI/GitHub Actions review initially found that rename detection could omit the source domain and that ordinary label additions could cancel full verification. The change disables rename collapsing and isolates label-event concurrency; the reviewer re-reviewed the correction and passed it with no remaining issues.


MrImmortal09 (approved):
LGTM.

PR #369, opened 2026-08-31
Add the fedimint-sdk crate: scaffolding and API skeleton

zeenix:
Implements the first part of #344: the crate scaffolding and the full public API, with no implementation behind it.

This follows the staged approach agreed in the thread — land the API we agree on with `unimplemented!()` bodies everywhere and no FFI wiring, then divide the implementation per module across contributors. The API here is the RFC **as amended by the review in the thread**, and then by the review rounds on this PR.

## What's here

A standalone crate at `rust/fedimint-sdk` — the full public surface, every effectful body `unimplemented!()`, unit tests for the pure parts, and its own CI workflow.

- `Sdk` owns one storage, one seed and N federations. `leave()` is split into `close_federation` (retain data) and `forget_federation` (destructive, guarded on zero balance and no pending or reclaimable value).
- `Federation` hands out capability facades as `Option`, plus `capabilities()`, so an app decides what to render before the user acts rather than by catching an error.
- One `Operation` model replaces the per-module subscribe/await pairs. Operations run detached and survive restart; `updates()` hands out independent subscribers; `Ok(None)` means only that a final state was observed. Failed payments are states, never errors.
- Lightning, ecash and on-chain sends are quote-then-send: `quote()` returns an opaque expiring plan that `send()` consumes. **The quote is the execution authorization.** It binds the amount, the aggregate fee and total debit, the route or destination, and the inventory or configuration it was computed against; `send()` executes exactly that or fails with no effect (`QuoteExpired`, `QuoteChanged`). Details records carry plain `fee` / `total_debited` / `net_credit` figures, and a receive credits an unsigned amount that never consumes existing balance.
- Every details record is persisted in the write that creates the operation, so an operation id alone is enough to rebuild a receipt, a QR code or a deposit screen after a restart.
- Recovery is part of the ordinary API: `Sdk::recover`, `resume_recovery` and `recovery_status`, a recovery lock that only a completed recovery releases, and a documented erase path for one that cannot be finished.
- `Meta` keeps the raw config and consensus sources separate underneath a merged convenience view; `activity()` is documented as local history, with amount, fee and direction as separate fields defined once for every kind.

[...]

Issue #344, opened 2026-08-20
RFC: high-level Rust crate API


zeenix:
*Generated by Claude (Fable 5) under the guidance of @zeenix.*

Proposal for `crates/fedimint-sdk`: one ergonomic Rust API over `fedimint-client`, and the single surface every binding generates from. `fedimint-sdk-ffi` becomes `#[uniffi::export]` over these types, `fedimint-sdk-wasm` becomes wasm-bindgen over the same types, and the TS/RN packages reduce to thin renames. All languages share one shape.

Status: draft for discussion. State enums and field lists are illustrative. The shapes, semantics, and boundaries are the proposal.

## What using it looks like

```rust
let sdk = Sdk::builder()
    .storage(Storage::at("/data/fedimint")?)
    .mnemonic(Mnemonic::from_str(words)?)
    .build()
    .await?;

// name, network, guardians, modules, meta
let preview = sdk.preview(&invite).await?;
let fed = sdk.join(&invite).await?;

let balance = fed.balance().await?;

// ecash
let send = fed.ecash().send(Amount::from_sats(500)).await?;
share(send.notes.to_string());
let op = fed.ecash().receive(&Notes::from_str(&pasted)?).await?;

// lightning
// fee + selected gateway
let quote = fed.lightning().quote(&invoice).await?;
let op = fed.lightning().send(&invoice).await?;
let recv = fed.lightning().receive(Amount::from_sats(1_000), "coffee").await?;
show_qr(recv.invoice.to_string());

// every flow returns the same Operation handle
let mut updates = op.updates();
while let Some(state) = updates.next().await {
    render(&state);
}

// after restart
let op = fed.operation(&op_id);
let page = fed.activity(None, 20).await?;
```

## Design rules

- The crate defines its own types. No `fedimint-*` type crosses this API. UniFFI forces this anyway, and it means SDK releases stop being coupled to the fedimint release train: the crate pins published `fedimint-*` versions and hops lines via `Cargo.toml`.
- Every public type is expressible in UniFFI: objects with `&self` async methods, plain records, data enums. No generics on the FFI boundary, no tuples, no borrowed returns. The wasm export mirrors the identical shape.
- No module plumbing in the API. No instance IDs, no `get_first_module::<T>()`, no module version detection in app code.
[...]

PR #374, opened 2026-09-04
build(sdk): track fedimint master, pinned to one revision

zeenix:
Follow-up to #369 (ref #344). Every implementation PR needs the fedimint client stack, so this one lands the dependency policy on its own.

The crate tracks fedimint `master` through git dependencies pinned to one revision, and the flake's `fedimint` and `fedimint-wasm` inputs move to the same revision. The client the SDK compiles against and the federation devimint starts for its tests can no longer disagree. Both pins are bumped together from now on, and a `pins-agree` CI job fails the build when `Cargo.toml`, `Cargo.lock`, `flake.nix` or `flake.lock` name a different repository or revision.

- the sixteen `fedimint-*` crates are declared and unused, the bodies stay `unimplemented!()`
- native CI installs `libclang` and `cmake` (`librocksdb-sys` and `aws-lc-sys` run bindgen) and gets a longer timeout, since the fedimint stack now compiles on every run
- `.cargo/config.toml` sets the `getrandom` wasm backend cfg for local wasm builds, mirroring fedimint's own build. CI passes without it because the toolchain action exports `RUSTFLAGS`
- the flake bump also moves fedimint's `nixpkgs`, so the wasm bundle and the JS suite rebuild against the new revision. `verify.yml`'s test job is where a breakage there would show

## Testing

- the five crate gates (fmt, clippy `-D warnings`, rustdoc `-D warnings`, tests, wasm check) locally at this revision
- `pins-agree` run against the branch and against copies with a moved revision, a different repository owner, a stale `Cargo.lock`, a stale `flake.lock`, one flake input moved, and an input block collapsed onto one line. each fails with a single `::error::` line

### DIFF
--- a/.github/workflows/rust-sdk-ci.yaml
+++ b/.github/workflows/rust-sdk-ci.yaml
@@ -21,8 +21,9 @@ on:
       - 'rust/fedimint-sdk/**'
       - '.github/workflows/rust-sdk-ci.yaml'
       # The crate pins the same fedimint revision the flake does, so a flake
-      # bump has to re-run this workflow: pins-agree below is what catches a
-      # bump that moved one of the two and not the other.
+      # bump has to re-run this workflow: pins-agree below checks flake.nix
+      # and flake.lock against the crate's pin, and catches a bump that
+      # moved one of the three and not the others.
       - 'flake.nix'
       - 'flake.lock'
   pull_request:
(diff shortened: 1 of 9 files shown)


MrImmortal09 (approved):
LGTM

PR #375, opened 2026-09-04
feat(sdk): parse and format the value types

zeenix:
Stacked on https://github.com/fedimint/fedimint-sdk/pull/374 (ref #344). First slice of the implementation task map: the value types.

Every opaque string type in `fedimint_sdk` now parses and prints for real instead of holding whatever string it was given. `InviteCode`, `FederationId`, `OperationId`, `GatewayId`, `Txid`, `Address`, `Bolt11Invoice`, `Notes`, `Preimage` and `Mnemonic` wrap the upstream or bitcoin type, validate on `FromStr`, print the canonical form on `Display`, and round-trip. The accessors the skeleton promised are real: an invite code's federation id, an invoice's amount, description and expiry, the value of a bundle of notes, a mnemonic's words and a fresh 12-word phrase from OS entropy. `Amount` and `Sats` parse exactly their `Display` form. `Cursor` is untouched until the activity facade lands.

- nothing upstream crosses the public API. constructors from upstream values and the network helpers the lightning and on-chain facades will need are `pub(crate)`
- parse errors for the bearer values (`Mnemonic`, `Notes`, `InviteCode`) are fixed strings that never echo the input. the other types append the upstream message
- `Mnemonic` has no `Display` or `Debug`, and zeroizes on drop through bip39's `zeroize` feature. `generate` maps a failed entropy draw to `ErrorCode::Entropy` instead of panicking
- `is_expired` reads the clock through fedimint's shim, so it works on wasm
- every test fixture that used to construct these types from an unvalidated string now parses a real literal (checksummed invite code, encoded ecash token, signed invoices)

## Testing

- unit tests per type: round-trip, whitespace and case handling, wrong length, overflow, redaction of `Debug`, fixed error messages


MrImmortal09 (dismissed):
LGTM.

MrImmortal09 (approved):
LGTM.

PR #376, opened 2026-09-04
feat(sdk): storage, the builder and the federation lifecycle

zeenix:
Stacked on https://github.com/fedimint/fedimint-sdk/pull/375 (ref #344). Storage, the builder and the federation lifecycle over `fedimint-client`. Together with the operation infrastructure that follows, this is the engine every facade builds on.

An `Sdk` now opens for real. One storage location holds one seed and every federation the user joins, each federation namespaced inside the same database by a fixed-width prefix. `SdkBuilder::build` takes the single-opener lock, reconciles the seed (writes one into empty storage, refuses a different mnemonic over an existing seed, refuses storage that holds state but no readable seed), finishes any erase a previous run committed, and brings every stored federation back up, quarantining the ones that fail instead of failing the build. `preview` and `join` download the config, enforce the module generation rule (a federation is all v1 or all v2 across mint, wallet and lightning, anything mixed is refused with the conflicting modules named), and derive network and capabilities. `close`, `reopen`, `forget` and `shutdown` follow the documented phases: quiescing in-flight calls, refusing to forget a federation with a balance or pending operations, tombstoning before erasing, announcing `Forgotten` once.

- storage backends: RocksDB on native, an OPFS-backed store on wasm (compiled, not yet exercised in a browser), memory for tests. The single-opener lock is an OS file lock, so a dead process never leaves a stuck lock
- the SDK's own records live under their own tag. The two stored enums have hand-written one-byte codecs because the upstream `Decodable` derive on enums expands to `anyhow` paths, which this crate does not have. fedimint/fedimint#9103 fixes the derive
- every facade call holds a read guard over the client for its whole duration. Closing a federation takes the write guard, so the wait is the quiesce
- a devimint harness lands with the first integration tests (join, read a zero balance, survive a restart, reopen replaces the handle, a mixed-generation federation is refused). The tests skip with a notice when devimint is absent and fail loudly with `FM_SDK_REQUIRE_DEVIMINT=1`. CI wiring for them is a later task

## Testing

- unit tests on in-memory and RocksDB storage: seed reconcile in all four cases, the lock and its reclaim, the generation rule and network derivation over hand-built module sets, close/reopen/forget state transitions with planted records
- the integration tests were compiled but not run against devimint on this machine. `scripts/run-sdk-integration-tests.sh v1|v2|mixed` is the entry point


MrImmortal09 (approved):
LGTM.

PR #377, opened 2026-09-05
feat(sdk): the operation infrastructure behind every facade

zeenix:
Follow-up to #376 (ref #344). The operation infrastructure every facade calls into, completing the engine the facades build on.

`Operation<S>` is real. Each facade creates an operation through one internal call that writes the SDK's own record and a chronological index entry, and hands back a typed handle. `state()` reads the current state through a per-kind driver, `details()` returns the typed record, `updates()` hands out independent subscribers that yield the current state first, then every transition, and end with `Ok(None)` once a final state was delivered. A subscriber parked on a stream resolves with `FederationClosed` when the federation stops. `AnyOperation` recovers a typed handle for a stored operation and answers whether this build can read it. `request_cancel` gets its durable half, a persisted request the ecash facade acts on.

- fedimint gives an embedder no way to write in the same transaction as a module's operation-log entry, so the log is authoritative and the SDK record follows it. A reconciliation pass on every federation open backfills a record for any log entry that lacks one. The upstream ask is fedimint/fedimint#9117
- upstream update streams end silently when a subscriber lags. The engine re-subscribes once and treats a second truncation as an error
- a final state is handed out only after it was persisted. If the caller drops the future mid-persist, the next call retries and delivers it
- drivers are a build-wide lookup keyed by operation kind, so a facade adds one match arm. No facade driver ships here, a scripted test driver exercises every clause of the contract
- three points differ from the written plan on purpose: a reconciliation rewrite only touches placeholder records (kind equal to module), so a record written by a newer build keeps its phase, cancellation intent and final state. reconciliation warns and continues past a bad entry instead of aborting. a record write checks for an existing record inside its transaction, so a fresh creation beats a placeholder and never a real record

## Testing

- unit tests over in-memory storage: record and index round-trips, every `updates()` clause (current first, dedupe, final then `None`, independent subscribers, cursor advances on hand-off, re-subscribe on truncation, release on close, handoff retry), the typed accessors, backfill from a directly written log entry, reattach after a RocksDB close and reopen


rust/fedimint-sdk/src/db.rs
```diff
@@ -442,21 +442,211 @@ fn backend_error(err: fedimint_core::db::DatabaseError) -> Error {
...
+pub(crate) struct OperationRecordKeyPrefix;
+
+/// What the SDK knows about one operation that the client's own log does not record.
+///
+/// The client's operation log is authoritative for an operation's existence and its module; this
+/// record adds the SDK's reading of it: which of the SDK's kinds it is, when it was created, the
+/// details record the creating call promised to persist, and the two facts that have to survive a
+/// restart because no state carries them (the phase a phase-keyed mapping reads, and whether a
+/// cancellation was asked for).
+#[derive(Debug, Clone, PartialEq, Eq, Encodable, Decodable)]
```
zeenix: 🤖 Chatgpt GPT-6 Astra Pro  on behalf of zeenix: **[P1] Redact persisted payloads from `Debug`.**

`AnyOperation` derives `Debug` through `AnyOperationInner → OperationInner → OperationRecord`, so `format!("{any:?}")` prints `details` verbatim. Ecash details contain bearer notes; even the no-driver backfill path preserves upstream `SpendOOB` metadata, including `oob_notes`. This bypasses `Notes`' redacted `Debug`, allowing ordinary logging/crash reports to expose spendable funds. Please redact the opaque payload fields (`details` and potentially secret-bearing `final_state`), and add a regression test that formats an `AnyOperation` containing a sentinel token and verifies it is absent.

rust/fedimint-sdk/src/operation.rs
```diff
@@ -332,12 +472,161 @@ impl<S: OperationState> OperationUpdates<S> {
...
+                // Reloaded rather than the cached `self.inner.record`: a resubscribe can follow a
+                // `phase` write the first subscription's driver made after this handle was
+                // created, and a driver that folds its starting point from `phase` needs that
+                // write to be visible or it resumes from the wrong place.
+                let record = self.inner.reload().await?;
+                let stream = self
+                    .driver
+                    .subscribe(&self.inner.federation, self.inner.id, &record)
+                    .await?;
+                self.stream = Some(stream);
```
zeenix: 🤖 Chatgpt GPT-6 Astra Pro  on behalf of zeenix: **[P2] Retain in-progress subscription setup across cancellation.**

The `subscribe()` future lives inside this particular `next()` call; only its completed stream is retained on the subscriber. A driver can register its upstream receiver and then await its initial snapshot. Cancelling `next()` there drops that receiver; the following `next()` starts a fresh subscription from the newer current state, losing intervening transitions despite the documented cancellation-safety guarantee. The existing fixtures all complete `subscribe()` immediately, so they miss this case. Keep the initialization future/state on `OperationUpdates` across cancelled waits, and test a driver that registers a receiver, yields during setup, and receives transitions before the caller retries.

rust/fedimint-sdk/src/operation.rs
```diff
@@ -332,12 +472,161 @@ impl<S: OperationState> OperationUpdates<S> {
...
+                    // than an `expect`, because the binding layer is built with
+                    // `panic = "abort"`.
+                    return Err(Error::new(
+                        ErrorCode::Internal,
+                        "this operation's subscription went missing",
+                    ));
+                };
+                tokio::select! {
+                    item = futures::StreamExt::next(stream) => item,
+                    _ = closed.changed() => {
```
zeenix: 🤖 Chatgpt GPT-6 Astra Pro  on behalf of zeenix: **[P2] Observe federation closure outside the established-stream wait too.**

Only `stream.next()` is raced against `closed.changed()`. If `next()` is pending in the earlier `driver.subscribe().await`, or in the empty-stream fallback's `driver.current().await`, closing the federation never polls this branch and the outstanding call can remain pending instead of returning `FederationClosed`. Final-state persistence is outside the race as well. Extend shutdown handling to these pending phases while preserving their cancellation state. A deterministic regression test is a driver whose `subscribe()` stays pending, followed by `set_status(Closed)`; the existing close test only parks after subscription setup has completed.

zeenix (commented):
🤖 Chatgpt GPT-6 Astra Pro  on behalf of zeenix: Reviewed `de0796b6`. The bearer-token leak through `Debug` should be fixed before merge. Two subscription cancellation/shutdown gaps are also noted inline, with regression-test scenarios. Static review only; I could not run Rust tests in this environment.

rust/fedimint-sdk/src/db.rs
zeenix: Fixed in 2c47984. `OperationRecord` now has a hand-written `Debug` that prints `details` and `final_state` as `<redacted>`, so every handle that derives through it (`AnyOperation`, `OperationInner`) is covered in one place. Two regression tests: one formats a record with sentinel payloads in both fields, the other formats an `AnyOperation`, the typed `Operation` recovered from it and its `OperationUpdates`, and checks the sentinel never appears.

rust/fedimint-sdk/src/operation.rs
zeenix: Not changed in code, docs tightened in 2c47984. The guarantee is about the subscriber's cursor, and before the first `next()` resolves there is no cursor yet: the subscription is established lazily by that first call, and it begins from the state the operation is in when it is established. A first call dropped mid-`subscribe` and retried is indistinguishable from a first call made later. The retry sees the current state then, and everything after it, which is the same "current first, then every transition" contract a subscription opened at that moment has. The re-establishment after a truncated stream is the same shape and already resumes from the current state by design.

Retaining an in-flight `subscribe` across cancellations would need the driver seam to hand out `'static` futures, so owned `Arc`s and an owned record instead of the borrows it takes now, for every facade driver. That is a larger change to the seam the facades build on than the case warrants, since no transition a subscriber was positioned on can be lost either way. The `next()` docs now say exactly when the no-lost-transition guarantee applies and what a dropped first call means.

rust/fedimint-sdk/src/operation.rs
zeenix: Fixed in 2c47984. Every wait in the loop now goes through one `until_closed` helper that races it against the federation's `closed` watch: the `reload` plus `subscribe` that establish the stream, the stream itself, and the `reload` plus `current` of the direct-read fallback. Two new tests park `next()` inside a driver whose `subscribe`, respectively `current`, never resolves, then close the federation, and check the call returns `FederationClosed`.

The final-state persist is deliberately left outside the race. It writes to the SDK's own storage, which a closing federation does not take away, and abandoning it would turn a state the caller is owed into a retry through `handoff` for no gain.

rust/fedimint-sdk/src/db.rs
zeenix: 🤖 Chatgpt GPT-6 Astra Pro  on behalf of zeenix: Verified at `2c47984`: both opaque payloads are redacted at `OperationRecord`, including through `AnyOperation`, and the added regression tests cover both paths. This addresses the security finding. Resolving.

rust/fedimint-sdk/src/operation.rs
zeenix: 🤖 Chatgpt GPT-6 Astra Pro  on behalf of zeenix: Verified at `2c47984`: setup and fallback reads now use the closure race, with regression tests for each pending driver call. I accept letting the local final-state write complete: it does not depend on the stopped federation, and preserves persist-before-delivery. Resolving this finding.

rust/fedimint-sdk/src/operation.rs
zeenix: 🤖 Chatgpt GPT-6 Astra Pro  on behalf of zeenix: Still outstanding at `2c47984`. The added exception weakens the no-lost-transition contract rather than fixing setup cancellation; `OperationUpdates` and `Operation::updates` still promise the stronger behavior. Once the driver has registered a receiver, dropping setup can discard real buffered transitions even though `last` is still `None`.

This does **not** require changing `Driver`'s borrowed interface. Store a boxed `'static` **outer** `async move` future on `OperationUpdates`, owning cloned `inner` and `driver` Arcs. Load the record inside that future and call the existing borrowed `subscribe` there. Race a mutable reference to the stored future against closure: cancelling `next()` then preserves setup, while actual federation closure can discard it deliberately.

Please retain the existing contract and add the paused-setup cancellation regression test described above. This is fixable entirely inside the SDK engine.

zeenix (commented):
🤖 Chatgpt GPT-6 Astra Pro  on behalf of zeenix: Rechecked `2c47984`. Resolved the payload-redaction and shutdown-wait findings. The setup-cancellation P2 remains: the docs now weaken the contract instead of preserving pending setup. Replied on the existing thread with an SDK-internal solution that does not change `Driver`'s interface. No additional defects found in this update. Static review; Rust tests were not run locally.

rust/fedimint-sdk/src/operation.rs
zeenix: Done in 3400dfc, along the lines you suggested. `OperationUpdates` now owns the establishment as one boxed `'static` future (`setup`) holding clones of the handle and driver Arcs; the record reload and the borrowed `subscribe` run inside it. `next()` polls a `&mut` to that future through the same close race as everything else, so a dropped call leaves it in place and the following call resumes it, while a federation that stops discards it. `Driver` is unchanged.

The docs are back to the unqualified guarantee, with a sentence saying setup is kept from the very first call. The regression test uses a driver whose `subscribe` takes its receiver, then parks at a gate until the test opens it: the test polls a `next` until the driver has registered, drops it, sends two transitions, opens the gate, and checks the following calls deliver both with exactly one registration. With the setup discarded on the drop instead, that test parks for ever at the gate on the second registration, which the bounded wait turns into a failure.

zeenix: 🤖 Chatgpt GPT-6 Astra Pro  on behalf of zeenix: Verified at `3400dfc`: `setup` owns the outer future, and `until_closed` polls it by mutable reference, so cancelling `next()` preserves the registered receiver without changing `Driver`. The regression test exercises cancellation after registration, buffers both transitions, and checks they arrive with exactly one registration. I also confirmed that test passed in CI. This addresses the remaining finding; resolving.

zeenix (commented):
🤖 Chatgpt GPT-6 Astra Pro  on behalf of zeenix: Rechecked `3400dfc`. All three findings are now resolved; no remaining blockers from my review. Pending setup survives a cancelled `next()` without changing `Driver`, and I confirmed the new regression test passed in CI (275 unit tests passed; fmt/clippy/docs/wasm also green). Ready to merge from my review. Local execution was unavailable because the container could not resolve the Rust download host.

Non-blocking for the facade follow-ups: add a cancelled-setup → close test using the real `ClientGuard`/`quiesce` path, not only detached drivers and `set_status`. A future driver must not leave a lifecycle read guard parked in an unpolled subscriber, since `quiesce` needs its write lock before signalling closure. No production driver ships in this PR.

rust/fedimint-sdk/src/operation.rs
zeenix: Your non-blocking note turned out to be blocking, so I took the retained setup out again in ee8f9dc.

I wrote the test you described first: a driver that holds the client read guard while parked inside `subscribe`, a `next()` driven until it is parked there and then dropped, then `quiesce()` under a timeout. With the establishment kept on the subscriber it hangs: the guard lives in a future nothing polls, and `quiesce` takes the write lock before it can signal. That is not a driver-discipline problem for the facades to avoid. Every real driver holds the guard across its operation-log read, so any `next()` dropped during that read left an idle subscriber that blocked the close.

The transition loss that retention was meant to prevent cannot happen with fedimint's drivers. `subscribe_spend_notes` and the rest await only the operation-log read and return a lazily built stream through `outcome_or_updates`; the notifier subscription runs inside the `stream!` body, on first poll, and the engine keeps that stream. A `next()` dropped before its stream exists has registered nothing. That shape is now the recorded obligation on `Driver::subscribe`, with the upstream citations, the `next()` docs state the unqualified guarantee and why it holds from the first call, and the close test is the regression test. `Driver` is unchanged.

MrImmortal09 (approved):
LGTM, I do have a few suggestions but lets have them in upcoming PRs.

PR #381, opened 2026-09-08
feat(sdk): the lightning facade, over either module generation (T8)

zeenix:
Follow-up to #377 (ref #344). The lightning facade, the first facade that moves money end to end on the operation engine.

`Lightning::quote`, `send` and `receive` run over both module generations behind one surface. A quote freezes what the user approves: the invoice amount, the gateway's charge, the federation's own fees for funding the contract and the dust, priced through the client's shared fee dry-run so the parts sum to the total exactly. Executing it reads the same inputs again and refuses on any change, and an invoice the module already paid is refused before anything is funded. A receive issues the invoice through a gateway the module selects, with the receive-side fee taken out of what lands, and persists everything a screen needs to redraw the QR code after a restart. Two drivers fold the four upstream state machines onto the two public state enums.

- v1's `Rejected` means two things depending on whether the receive was ever funded, so the phase is persisted and the answer survives a restart. lnv2's `Failure` is the same story on the send side
- a claim the federation rejects on v1 is retried under the same SDK operation. The module can only retry as a new upstream operation, so the retry's id is recorded and followed, one retry per receive, started as a task a dropped poll cannot cancel
- a quote is executed exactly or not at all, with one exception. When the module settles a gateway-quoted v1 payment inside the federation after the route re-check, the record backs out the gateway charge it knows was not taken and says the rest is the quote's estimate
- `Bolt11Invoice` equality goes by the bolt11 encoding. An invoice a module builds and the same invoice parsed back differ in fields BOLT11 does not carry
- the quoted terms ride along in the module's own metadata, so a record rebuilt from the module's log after a crash carries the exact fees, not an estimate
- lnv2 payments and invoices are funded at the terms the SDK checked, through the `send_with_terms` and `receive_with_terms` calls from fedimint/fedimint#9125. Until that lands on master the fedimint pin follows the `rust-sdk` branch upstream, whose head is that PR's tree
- two places where the diff carries an upstream limitation instead of hiding it
  - v1 hands back a preimage two characters short (fedimint/fedimint#8969), so the v1 send test asserts `Internal` where the lnv2 one asserts `Success`
  - neither module pays a `tb` invoice on testnet4 (fedimint/fedimint#9100), surfaced as `NetworkMismatch` and noted in the error docs
- the engine gains the two things the drivers needed: record writers a stream can call, and the creation time on the backfiller

## Testing

- unit tests for every mapping table, the wire records, the quote checks and arithmetic, and the drivers through the engine on a detached federation
- the devimint suite on both shapes, 8/8 each: an issued invoice paid from outside, with the credit landing and the operation readable by id before and after a restart; an outbound payment that debits exactly the quoted total and ends with a preimage on lnv2, while on v1 the test asserts the #8969 outcome; the refusals with their details; a quote used twice
- on the lnv2 shape the harness pins the guardian's gateway list to the LND gateway, because the lnv2 client shuffles gateways on purpose and `wasm-test-setup` funds only that one

### DIFF
--- a/rust/fedimint-sdk/src/lightning/driver.rs
+++ b/rust/fedimint-sdk/src/lightning/driver.rs
@@ -1,12 +1,18 @@
 //! The two lightning drivers, the backfiller, and the stream helpers they share.
 
 use core::time::Duration;
+use std::any::Any;
 
+use fedimint_core::core::OperationId;
 use fedimint_core::task::MaybeSend;
-use fedimint_core::util::BoxStream;
+use fedimint_core::util::{BoxFuture, BoxStream};
 use futures::{Stream, StreamExt, future};
 
-use crate::{Error, ErrorCode, OperationState, Result};
+use super::{v1, v2, wire};
+use crate::db::OperationRecord;
+use crate::federation::FederationInner;
+use crate::operation::{Backfilled, Backfiller, Driver};
+use crate::{Error, ErrorCode, LnReceiveState, LnSendState, OperationState, Result};
 
 /// How long `current` waits for the next replayed state before calling the last one current.
 ///
@@ -60,4 +66,125 @@ where
             "this operation's subscription yielded no state",
         )
     })
+}
+
+/// Observes an outgoing lightning payment of either generation, chosen by the record's module.
+pub(crate) struct LnSendDriver;
+
+impl Driver<LnSendState> for LnSendDriver {
+    fn current<'a>(
+        &'a self,
+        federation: &'a FederationInner,
+        id: OperationId,
+        record: &'a OperationRecord,
+    ) -> BoxFuture<'a, Result<LnSendState>> {
+        Box::pin(async move {
+            if let Some(encoded) = &record.final_state {
+                return wire::decode_send_state(encoded);
+            }
+            settle(self.subscribe(federation, id, record).await?).await
+        })
+    }
+
+    fn subscribe<'a>(
+        &'a self,
+        federation: &'a FederationInner,
+        id: OperationId,
+        record: &'a OperationRecord,
+    ) -> BoxFuture<'a, Result<BoxStream<'static, Result<LnSendState>>>> {
+        Box::pin(async move {
+            let details = wire::decode_send_details(&record.details)?;
+            match record.module.as_str() {
+                "ln" => v1::subscribe_send(federation, id, &details).await,
+                "lnv2" => {
+                    v2::subscribe_send(federation, id, record.phase.unwrap_or(0), &details).await
+                }
+                other => Err(unknown_module(other)),
+            }
+        })
+    }
+
+    fn same_state(&self, previous: &LnSendState, next: &LnSendState) -> bool {
+        previous == next
+    }
+
+    fn encode_state(&self, state: &LnSendState) -> Result<String> {
+        wire::encode_send_state(state)
+    }
+
+    fn decode_details(&self, json: &str) -> Result<Box<dyn Any + Send + Sync>> {
+        Ok(Box::new(wire::decode_send_details(json)?))
+    }
+}
+
+/// Observes an incoming lightning payment of either generation.
+pub(crate) struct LnReceiveDriver;
+
+impl Driver<LnReceiveState> for LnReceiveDriver {
+    fn current<'a>(
+        &'a self,
+        federation: &'a FederationInner,
+        id: OperationId,
+        record: &'a OperationRecord,
+    ) -> BoxFuture<'a, Result<LnReceiveState>> {
+        Box::pin(async move {
+            if let Some(encoded) = &record.final_state {
+                return wire::decode_receive_state(encoded);
+            }
+            settle(self.subscribe(federation, id, record).await?).await
+        })
+    }
+
+    fn subscribe<'a>(
+        &'a self,
+        federation: &'a FederationInner,
+        id: OperationId,
+        record: &'a OperationRecord,
+    ) -> BoxFuture<'a, Result<BoxStream<'static, Result<LnReceiveState>>>> {
+        Box::pin(async move {
+            let phase = record.phase.unwrap_or(0);
+            match record.module.as_str() {
+                "ln" => v1::subscribe_receive(federation, id, phase, &record.details).await,
+                "lnv2" => v2::subscribe_receive(federation, id, phase).await,
+                other => Err(unknown_module(other)),
+            }
+        })
+    }
+
+    fn same_state(&self, previous: &LnReceiveState, next: &LnReceiveState) -> bool {
+        previous == next
+    }
+
+    fn encode_state(&self, state: &LnReceiveState) -> Result<String> {
+        wire::encode_receive_state(state)
+    }
+
+    fn decode_details(&self, json: &str) -> Result<Box<dyn Any + Send + Sync>> {
+        Ok(Box::new(wire::decode_receive_details(json)?))
+    }
+}
+
+/// Rebuilds a lightning record from the module's own log entry, for either generation.
+pub(crate) struct LnBackfiller;
+
+impl Backfiller for LnBackfiller {
+    fn backfill(
+        &self,
+        module_kind: &str,
+        meta: &serde_json::Value,
+        created_at: u64,
+    ) -> Option<Backfilled> {
+        match module_kind {
+            "ln" => v1::backfill(meta, created_at),
+            "lnv2" => v2::backfill(meta, created_at),
+            _ => None,
+        }
+    }
+}
+
+fn unknown_module(module: &str) -> Error {
+    Error::new(
+        ErrorCode::Internal,
+        format!("a lightning record names a module this build cannot observe: {module:?}"),
+    )
 }
--- a/rust/fedimint-sdk/src/lightning/driver.rs
+++ b/rust/fedimint-sdk/src/lightning/driver.rs
@@ -1,0 +1,63 @@
+//! The two lightning drivers, the backfiller, and the stream helpers they share.
+
+use core::time::Duration;
+
+use fedimint_core::task::MaybeSend;
+use fedimint_core::util::BoxStream;
+use futures::{Stream, StreamExt, future};
+
+use crate::{Error, ErrorCode, OperationState, Result};
+
+/// How long `current` waits for the next replayed state before calling the last one current.
+///
+/// Neither upstream module offers a "current state" read: v1's `subscribe_*` are generators that
+/// re-run from the first state and resolve each already-passed stage immediately, and lnv2's
+/// notifier replays the stored states before streaming new ones. Draining with a short wait per
+/// item is how a point-in-time answer is produced from that.
+const CURRENT_STATE_SETTLE: Duration = Duration::from_millis(500);
+
+/// Ends a stream after its first final state, and on the first error.
+pub(super) fn until_final<S>(
+    stream: impl Stream<Item = Result<S>> + MaybeSend + 'static,
+) -> BoxStream<'static, Result<S>>
+where
+    S: OperationState,
+{
+    Box::pin(stream.scan(false, |done, item| {
+        if *done {
+            return future::ready(None);
+        }
+        *done = match &item {
+            Ok(state) => state.is_final(),
+            Err(_) => true,
+        };
+        future::ready(Some(item))
+    }))
+}
+
+/// The last state a fresh subscription yields promptly: the current one.
+pub(super) async fn settle<S>(mut stream: BoxStream<'static, Result<S>>) -> Result<S>
+where
+    S: OperationState,
+{
+    let mut last = None;
+    loop {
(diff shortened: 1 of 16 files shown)


rust/fedimint-sdk/src/lightning/v2.rs
```diff
@@ -0,0 +1,739 @@
...
+    if current != (send_fee, expiration_delta) || fresh.total != quote.plan.total {
+        return Err(quote_changed(quote.plan.total, fresh.total));
+    }
+    let id = module
+        .send(
+            quote.invoice.inner().clone(),
+            Some(gateway),
+            serde_json::Value::Null,
+        )
+        .await
```
zeenix: 🤖 Chatgpt on behalf of zeenix: **P1 — Enforce the approved quote within funding.** After this recheck, upstream `send()` fetches gateway terms again and independently selects funding notes. Changed fees or concurrent balance changes can therefore produce a different debit, while lines 342–344 still record the approved quote. Pass binding terms into an upstream funding API that rejects drift atomically; another preliminary check cannot enforce this guarantee.

rust/fedimint-sdk/src/lightning/v1.rs
```diff
@@ -0,0 +1,1267 @@
...
+            Ok(id) => id,
+            Err(err) => {
+                return classify_reclaim_refusal(&err.to_string()).map(ReclaimStart::Impossible);
+            }
+        };
+        // Persisted before it is followed: after a restart the record is the only thing that
+        // says which upstream operation to follow.
+        details.reclaim_operation_id =
+            Some(crate::OperationId::from_upstream(reclaim_id).to_string());
+        write_details_in(&self.db, self.id, wire::encode_receive_wire(&details)?).await?;
```
zeenix, line 325: 🤖 Chatgpt on behalf of zeenix: **P1 — Recover an existing reclaim before starting another.** Upstream commits the retry before this write. A crash in between loses its ID, so reopening starts another retry; that retry can fail because the first already claimed the funds, incorrectly ending the SDK operation as `Failed`. The spawned task protects against a dropped poll, not a process crash. Recover the existing retry through upstream's `ReceiveReclaim.original_operation_id` metadata.

rust/fedimint-sdk/src/lightning/driver.rs
```diff
@@ -0,0 +1,190 @@
...
+        &'a self,
+        federation: &'a FederationInner,
+        id: OperationId,
+        record: &'a OperationRecord,
+    ) -> BoxFuture<'a, Result<BoxStream<'static, Result<LnReceiveState>>>> {
+        Box::pin(async move {
+            let phase = record.phase.unwrap_or(0);
+            match record.module.as_str() {
+                "ln" => v1::subscribe_receive(federation, id, phase, &record.details).await,
+                "lnv2" => v2::subscribe_receive(federation, id, phase).await,
```
zeenix, line 174: 🤖 Chatgpt on behalf of zeenix: **P2 — Start subscriptions at the current state.** Both drivers forward upstream replay streams directly; `settle()` only serves `state()`. A late subscriber without a cached upstream outcome therefore receives historical `Created`/`Funding` states, contrary to `Operation::updates()`'s current-state-first contract. Establish the current state and retain subsequent transitions without replaying the earlier history; cover late subscription before any upstream outcome has been cached.

rust/fedimint-sdk/src/lightning/v2.rs
```diff
@@ -0,0 +1,739 @@
...
+    )
+    .await?;
+    if current != (send_fee, expiration_delta) || fresh.total != quote.plan.total {
+        return Err(quote_changed(quote.plan.total, fresh.total));
+    }
+    let id = module
+        .send(
+            quote.invoice.inner().clone(),
+            Some(gateway),
+            serde_json::Value::Null,
```
zeenix: 🤖 Chatgpt on behalf of zeenix: **P2 — Preserve quoted terms in upstream custom metadata.** All four creation paths pass null metadata and write SDK details only after upstream commits. A crash between those commits permanently replaces exact fees and amounts with backfill estimates. Both upstream generations already persist custom metadata atomically with operation creation: store versioned SDK terms there and have the backfillers recover them. This loss is avoidable in the SDK, rather than an upstream limitation.

zeenix (commented):
🤖 Chatgpt on behalf of zeenix: Reviewed `0247b3b`: not ready to merge; four findings below. The exact-quote guarantee needs enforcement within upstream funding, while the remaining findings concern SDK recovery and subscription behavior.

rust/fedimint-sdk/src/lightning/v2.rs
zeenix: Upstream's `send` has no input for terms the caller already checked: it calls `routing_info` itself and funds `send_fee.add_to(amount)` from that read, so the SDK cannot make funding refuse drift. What it can do is make the record true. After `send` returns, the committed contract amount is read back from the operation log and, if the gateway's fee moved between the SDK's check and upstream's read, the record's fee and total are corrected to what was actually taken. A failed read keeps the quoted figures rather than reporting a started payment as failed. The `Lightning::send` docs now state the guarantee in those terms. The binding-terms API is an upstream feature request.

rust/fedimint-sdk/src/lightning/v1.rs
zeenix: Fixed. Before starting a retry, `start` now walks the client's chronological operation index newest first, down to the receive's own creation time, and follows any `ReceiveReclaim` entry whose `original_operation_id` is this receive. The found id is persisted on the record before it is followed, same as a freshly started one. Three unit tests cover found, older-than-the-receive (walk stopped) and a reclaim of a different receive.

rust/fedimint-sdk/src/lightning/driver.rs
zeenix: Fixed. Both drivers now wrap the upstream stream in `settled`, which awaits the first replayed state, drains whatever follows within the settle window, yields the last one, and forwards every later transition as it comes. `current` is the first item of that same stream. Four unit tests: late subscriber sees only the settled state, later transitions forwarded, replay ending in a final state yields only that state, empty stream yields nothing.

rust/fedimint-sdk/src/lightning/v2.rs
zeenix: Fixed. All four creation paths now pass the SDK's own details record under a `fedimint_sdk` key in the module's custom metadata, built before the upstream call. Both backfillers read it back when present and the invoice matches, and fall back to the estimate otherwise. The receive copy leaves the invoice and expiry to upstream's meta, since they are only known after the call. Docs on the details types now say a record is exact for an operation this SDK created and an estimate only for an entry it did not create.

rust/fedimint-sdk/src/lightning/v2.rs
zeenix: The binding-terms API is now requested upstream as fedimint/fedimint#9124.

rust/fedimint-sdk/src/lightning/v1.rs
zeenix: 🤖 Chatgpt on behalf of zeenix: Verified at `9051850`: the upstream log is searched for this receive's existing reclaim before another is started, and the recovered ID is persisted and followed. This addresses the crash window.

rust/fedimint-sdk/src/lightning/driver.rs
zeenix: 🤖 Chatgpt on behalf of zeenix: Verified at `9051850`: both drivers now drain the initial replay before yielding and retain the same stream for subsequent transitions. The added tests cover late subscription and forwarding. Resolving.

rust/fedimint-sdk/src/lightning/v2.rs
zeenix: 🤖 Chatgpt on behalf of zeenix: The upstream gap is now tracked in fedimint/fedimint#9124, but the replacement is still incorrect: `committed_fee` adjusts only the gateway component. lnv2's own consensus fee depends on the contract amount, and funding-note fees/dust can change too, so these are not the actual debit figures the new docs promise. Crash backfill also restores the uncorrected metadata copy.

Please revert this after-the-fact adjustment and the relaxation of the exact-quote API contract, keeping the temporary upstream limitation explicit with a linked TODO/regression test. This follows the established SDK direction: fix the missing guarantee upstream rather than redesign the app-facing contract around it. #9124 should also cover binding the funding debit/note selection, not just gateway terms. The upstream implementation itself need not block this PR; the inaccurate SDK workaround should be removed.

rust/fedimint-sdk/src/lightning/v2.rs
zeenix: 🤖 Chatgpt on behalf of zeenix: The metadata now preserves the terms, but introduces a recovery regression in `v1::backfill`: it also trusts `copy.route`. In the explicitly supported gateway-quoted → internally settled case, that copy still says `Gateway` while upstream's `pay.is_internal_payment` is true. After a crash before the SDK record write, `state()`/`updates()` select `subscribe_ln_pay`, which explicitly rejects internal payments, so the operation becomes unobservable.

Always derive the executed v1 route from upstream's payment metadata, even when the fee/total come from the SDK copy. Add a regression test with a gateway-route copy and `is_internal_payment: true`. Leaving this thread open for that fix.

rust/fedimint-sdk/src/lightning/v2.rs
zeenix: 🤖 Chatgpt on behalf of zeenix: Rechecked with fedimint/fedimint#9125 (`1ab15e4`). Its new `send_with_terms`/`receive_with_terms` bind the gateway fees used in the contract, addressing that part of #9124. They do not bind funding-note selection, mint fees or dust, so the full-debit guarantee remains a separate upstream gap.

This SDK PR is still at `9051850`, with the old dependency pin and `send`/`receive` calls. The requested removal of the inaccurate fee adjustment/API relaxation and the v1 backfill route fix therefore remain outstanding. #9125 provides the gateway-terms API to adopt once available; it does not fix those SDK changes by itself.

zeenix: Now enforced within funding. The pin follows the upstream `rust-sdk` branch, whose head is fedimint/fedimint#9125, and the lnv2 paths call `send_with_terms` and `receive_with_terms` with the terms the SDK checked. Upstream refuses before funding anything when the gateway's terms moved, which the send side reports as `QuoteChanged`. The post-commit correction is gone and the record is the quote again. On the receive side, where the fee is read an instant earlier in the same call, a changed fee is retried once at the fresh fee and a second refusal is reported as the gateway being unavailable.

rust/fedimint-sdk/src/lightning/v2.rs
zeenix: Fixed. A rebuilt v1 send now takes its route from the module's own record, never from the SDK copy. For the gateway-quoted, internally settled case the copy carries the gateway's share of the fee, so the rebuilt record backs it out the same way the live path does; a copy from an older build without that field keeps its figures as an upper bound, and a copy quoted internal is left alone. Regression tests cover a gateway-route copy with the module reporting an internal payment, both with and without the gateway share, and a gateway-route copy the module confirms.

rust/fedimint-sdk/src/lightning/v2.rs
zeenix: Two more things from this thread. The after-the-fact fee adjustment and the doc relaxation are gone with the checked-terms change, and the send docs state the exact-or-refused guarantee again. The remaining gap, that funding-note selection, mint fees and dust are not bound, is noted in a comment at the re-quote in the lnv2 send path and added to fedimint/fedimint#9124 as its second half.

zeenix: 🤖 Chatgpt on behalf of zeenix: Verified at `9ca15ce`: both lnv2 paths use #9125's checked-terms APIs; the post-commit adjustment and relaxed contract are removed. The receive retry recomputes the fee and metadata before retrying. The remaining funding-note/mint-fee/dust guarantee is explicitly tracked in fedimint/fedimint#9124, so it is deferred upstream rather than blocking this SDK PR. Resolving.

rust/fedimint-sdk/src/lightning/v2.rs
zeenix: 🤖 Chatgpt on behalf of zeenix: Verified at `9ca15ce`: v1 backfill always derives the route from upstream metadata. The optional gateway-fee field lets recovery match the live internal-settlement adjustment, while older records remain readable. Regression tests cover both cases. Resolving.

zeenix (commented):
🤖 Chatgpt on behalf of zeenix: Rechecked `9ca15ce`: both remaining SDK findings are addressed; all four review threads are resolved. No new blocking regressions found, and CI is green. Ready from this review, with the remaining funding-note/mint-fee/dust binding work deferred to fedimint/fedimint#9124.

MrImmortal09 (approved):
Let’s let the money roll in !! 🚀

Issue #380, opened 2026-09-08
T4 FFI Shape Spike


Ayush4958:
## T4 FFI Shape Spike: Separate Crate (Option A)

I prototyped both FFI options on the `spike/ffi-shape` branch. The recommendation is **Option A: a separate `fedimint-sdk-ffi` crate**.

### Why

The simplest reason: it's what the project already does. `fedimint-client-uniffi` is a standalone crate with its own lockfile and lint policy. Following that pattern for the new SDK requires no justification. Breaking from it does.

Option B (adding an `ffi` feature inside `fedimint-sdk`) would technically work under today's CI, but it creates three maintenance problems that Option A avoids entirely:

1. **Wasm fragility.** UniFFI generates `Send + Sync` bounds that don't compile on `wasm32-unknown-unknown` without an unstable feature flag. Today's CI doesn't pass `--all-features` to the wasm check, so it wouldn't break right now. But any contributor who runs `cargo check --all-features --target wasm32-unknown-unknown` locally will hit a wall, and any future CI tightening would surface it.

2. **Lint suppression.** The SDK enforces `#![deny(missing_docs)]`. UniFFI's `setup_scaffolding!()` macro generates undocumented internal items. In Option B you'd need to conditionally suppress that lint behind `cfg(feature = "ffi")`. The existing FFI crate sidesteps this by simply not setting the lint.

3. **Lockfile bloat.** Even as an optional dep, UniFFI's proc-macro tree (askama, pest, cargo_metadata, etc.) lands in the SDK's committed `Cargo.lock`. Every contributor resolves it whether they touch FFI or not. A separate crate keeps its own lockfile, same as `fedimint-client-uniffi` does today.

### What the prototype covers

- Wraps `Mnemonic::generate()` and `Mnemonic::from_phrase()` with `Arc<Self>` ownership
- Maps the SDK's `Error` / `ErrorCode` into a flat `FfiError` enum for the binding layer
- Matches uniffi 0.29, proc-macro approach, and `["staticlib", "cdylib", "lib"]` crate-type from the existing FFI crate
- Includes unit tests for generate, round-trip parse, and error rejection

**Note :-** This thread content is written by AI but It passes the manual review of author

zeenix:
@Ayush4958 please create PR for this, rather than issue and disconnected branch. If you're unsure of the approach, please make the PR draft.

Ayush4958:
[...]

PR #382, opened 2026-09-09
feat(sdk-ffi): scaffold uniffi bindings crate (T4 spike)

Ayush4958:
Adds `rust/fedimint-sdk-ffi/`, a standalone UniFFI bindings crate wrapping
`fedimint-sdk`. This is the T4 FFI shape spike: prototype one method both ways,
Decision Reasoning :- #380 

## What this PR adds

- `Cargo.toml` pinned to uniffi 0.29, matching `fedimint-client-uniffi`
- `src/lib.rs` wrapping `Mnemonic::generate()` and `Mnemonic::from_phrase()`
  with `Arc<Self>` ownership and `FfiError` error mapping
- `uniffi-bindgen.rs` CLI entry point
- `README.md` explaining why a separate crate
- `.gitignore` matching the existing FFI crate

## Decision: Option A (Separate Crate)

A separate crate is the recommended path over an in-crate `ffi` feature because:

1. The SDK enforces `#![deny(missing_docs)]`. UniFFI's `setup_scaffolding!()`
   generates undocumented internal items that break this lint.
2. UniFFI generates `Send + Sync` bounds incompatible with `wasm32-unknown-unknown`
   without unstable flags. Keeping it out of the SDK avoids wasm fragility.
3. A separate lockfile keeps UniFFI's proc-macro dependency tree out of the SDK's
   `Cargo.lock`, matching how `fedimint-client-uniffi` already works.

## What this does NOT include

- No full bindings. This is the shape decision only.
- No changes to `nix/ffi.nix` or `ubrn.config.yaml` (those come at cutover).
- No changes to `fedimint-sdk` itself.

**Note :-** This PR is created by AI but It goes through the manual review of author

### DIFF
--- a/js/.changeset/ffi-shape-spike.md
+++ b/js/.changeset/ffi-shape-spike.md
@@ -1,0 +1,5 @@
+---
+'@fedimint/core': patch
+---
+
+Add fedimint-sdk-ffi crate scaffold: T4 FFI shape spike prototyping separate-crate UniFFI bindings for Mnemonic.
--- a/rust/fedimint-sdk-ffi/.gitignore
+++ b/rust/fedimint-sdk-ffi/.gitignore
@@ -1,0 +1,38 @@
+# Rust related
+target
+build
+
+# Nix
+result
+result-*
+
+# Kotlin related
+.gradle
+wallet_db
+local.properties
+*.log
+*.dylib
+*.so
+*.db
+*/data/signet
+.DS_Store
+testdb
+.lsp
+.clj-kondo
+.idea/
+.editorconfig
+.kotlin/
+
+# Swift related
+/.build
+.swiftpm
+/Packages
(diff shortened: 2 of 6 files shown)


MrImmortal09 (commented):
We would have to make 2 changes if someone wants to add a feature to fedimint sdk, one at fedimint/fedimint side then one at fedimint-sdk/rust/fedimint-sdk-ffi. Thats the only reason I am trying to avoid this. 

We could have a seperate crate just to contain the bindgen but still have the code in the rust/fedimint-sdk itself

edit: 2 changes.

js/.changeset/ffi-shape-spike.md
MrImmortal09, line 1: we don't need a core patch, We need it only when we are any files linked to the js we have or packages/core to be particular.

PR #387, opened 2026-09-10
feat(sdk): activity history over the operation log (T11)

zeenix:
## Description

- ref fedimint-sdk-implementation-plan T11; depends on nothing still open, lightning (#381) supplies the real rows
- `Federation::activity` pages over the SDK's own operation records newest first, one row per record, with the amount, fee, direction and outcome bucket the rustdoc on `ActivityItem` and `ActivityStatus` fixes. Both tables are implemented once, in `src/activity/rows.rs`, one impl per state enum and per details record, so the ecash, on-chain and recovery rows are written and unit-tested now and become reachable when each facade lands its driver arm
- a row whose ending was never observed is brought up to date on the way out through the same read `Operation::state` does, so a receive paid while the app was closed lists as done rather than pending. Those reads run concurrently per page. A refresh that fails with `FederationClosed` or `Storage` fails the call; any other failure leaves the row at what its record says, which is what a v1 send hits today because of fedimint/fedimint#8969
- every record renders, including the ones this build cannot open: an unrecognised kind, a state schema newer than this build reads, or a recorded ending it cannot decode all report `Unknown` with no numbers, and finality comes off the record independently of the outcome
- `Cursor` is now the position after the last row, bound to the federation that issued it, rendered as opaque hex; a cursor from another federation and a zero limit are `InvalidInput`
- `Driver::decode_state` joins `encode_state`; a driver added later (the ecash one in #383) has to implement it too

## Testing

- 384 unit tests (35 new: the two tables variant by variant, the cursor, the page walk with ties, cursors, unknown kinds, newer schemas, undecodable endings, refresh and refresh failure)
- devimint 9/9 on both the v1 and v2 shapes: the new test lists a receive and a send, pages through them with a cursor that survives a string round trip, and reads the same rows after a restart

### DIFF
--- a/rust/fedimint-sdk/src/activity/page.rs
+++ b/rust/fedimint-sdk/src/activity/page.rs
@@ -207,8 +207,10 @@ where
                 return Err(err);
             }
             // An upstream state this build's driver cannot decode (fedimint/fedimint#8969 on
-            // the v1 module today) must not make the whole list unreadable: the record still
-            // says what it knows.
+            // the v1 module today) must not make the whole list unreadable, and it is not a
+            // reading either: `Pending` would claim the operation is still running when all
+            // that is known is that no ending was recorded. So the outcome is `Unknown` and
+            // finality is what the record says, which in this branch is "not yet".
             Err(err) => {
                 tracing::warn!(
                     target: "fedimint_sdk",
@@ -216,7 +218,7 @@ where
                     error = %err,
                     "could not refresh this operation's current state",
                 );
-                (ActivityStatus::Pending, false)
+                (ActivityStatus::Unknown, false)
             }
         },
     };
@@ -789,7 +791,7 @@ mod tests {
     }
 
     #[tokio::test(flavor = "multi_thread")]
-    async fn a_refresh_that_fails_leaves_the_row_at_what_its_record_says() {
+    async fn a_refresh_that_fails_is_an_unknown_outcome_with_the_recorded_finality() {
         let db = federation_namespace(&in_memory_root(), [1u8; 32]);
         let federation = FederationInner::detached(db.clone(), true);
         // `Operation::state` reloads the record before calling the driver, so the id it is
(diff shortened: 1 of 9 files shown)


zeenix (commented):

Reviewed `cbf13ef`. One correctness issue to fix before merge: an unreadable current state is reported as pending, contrary to the existing activity contract. The accounting/status tables, cursor ordering and federation binding, and persisted-final-state path otherwise look sound. CI is green.

rust/fedimint-sdk/src/activity/page.rs
```diff
@@ -0,0 +1,839 @@
...
+            // the v1 module today) must not make the whole list unreadable: the record still
+            // says what it knows.
+            Err(err) => {
+                tracing::warn!(
+                    target: "fedimint_sdk",
+                    operation = %inner.id.fmt_full(),
+                    error = %err,
+                    "could not refresh this operation's current state",
+                );
+                (ActivityStatus::Pending, false)
```
zeenix: 🤖 ChatGPT (Codex) on behalf of zeenix:

**[P2] Report an undecodable current state as `Unknown`, not `Pending`.**

`final_state == None` only means the SDK has not recorded an ending; it does not establish that the operation is still running. For example, the successful v1 send affected by fedimint/fedimint#8969 reaches this branch and is displayed as pending forever, including after restart. `ActivityStatus::Pending` explicitly excludes states this build cannot interpret.

For an undecodable state, return `Unknown` with the recorded finality; the existing figures-clearing code then enforces the public contract. Update the failing-driver test and the v1 integration expectations accordingly. The upstream decoding bug can remain upstream; the activity row should accurately report that its outcome is unknown.

zeenix: Fixed in cb4733e. A refresh that fails for anything other than a closed federation or a storage failure now yields `Unknown` with the recorded finality (`is_final == false` in that branch), so the figures are cleared by the existing rule. The `Unknown` docs name this as the third situation. The unit test and the v1 branch of the devimint test assert the new answer; devimint v1 passes 9/9.

zeenix: 🤖 ChatGPT (Codex) on behalf of zeenix:

Verified in `cb4733e`: unreadable current states now yield `Unknown` with no figures; recorded finality and storage/closed error propagation are preserved. The docs and unit/integration expectations match. No new issues found in the update, and CI is green. Ready to merge from my review.

omswami2004 (approved):
LGTM! Verified and approved.

> *Generated on behalf of [Om Swami](https://github.com/MrImmortal09) by AI*

MrImmortal09 (approved):
LGTM

PR #388, opened 2026-09-10
feat(android): init android SDK and expand uniffi coverage (T15)

MrImmortal09:
## Summary

Adds an Android SDK: the `android/` Gradle project whose Kotlin bindings are generated from `rust/fedimint-sdk`'s `uniffi` feature, the cacheable Nix cross-compile that produces its native libraries, the CI that builds the native library once and generates bindings from that artifact, and a demo app that drives every exported call from one screen.

**Base branch note:** this PR was originally stacked on `om/ffi-stage1`. That branch's own commits (T4, T7, T11, T12) have since landed on `main` individually, so the branch is now rebased directly onto current `main` — no merge commits, every commit here is this PR's own.

## Details

There is no hand-written Kotlin. The `#[uniffi::export]` blocks in `rust/fedimint-sdk` hand out that crate's own types (`Sdk`, `Federation`, `Ecash`, `Lightning`, `Onchain`, `AnyOperation`, `Mnemonic`, `InviteCode`, `Notes`, and the rest), so this SDK is a *view* of that API rather than a copy that could drift from it. The surface now covers the whole per-federation API: federation lifecycle (status, reopen, close, forget, shutdown), the `ecash`/`lightning`/`onchain` facades' `quote` → `send`/`receive` calls, operation lookup and typed downcasts, seed recovery, activity history and metadata — not just the three `Sdk` bootstrap methods (`createFedimintSdk`, `preview`, `join`) the feature started with.

Two pieces of shared machinery live in the new `ffi.rs`:

- **`ffi_operation!`** monomorphises one UniFFI object per concrete `Operation<S>` instantiation, since the real type is generic and a UniFFI object cannot be. Invoked once per facade, next to the state type it wraps.
- **`QuoteClaim`** is the single-use guard every quote type (`EcashQuote`, `LnQuote`, `OnchainQuote`) carries. In plain Rust, `send` takes its quote by value, so a second attempt with the same quote is a compile error; a Kotlin/Swift binding only ever holds a shared `Arc`, so nothing else stops it from calling `send` twice. `QuoteClaim` turns that into a runtime `QuoteExpired` (`already_executed: true`) on the second attempt, shared across all three facades instead of copied three times.

`Notes` (ecash) crosses as a UniFFI object, the same way `InviteCode` and `Mnemonic` already do, rather than a bare `String`. Ecash notes are a bearer instrument — printing one hands over spendable value — and a generated Kotlin/Swift record prints every field through its automatic `toString()`. As an object, the token comes out only through an explicit `display()`. The two Rust records that hold notes by value (`EcashSendDetails`, `EcashReceiveDetails`) keep their plain-Rust shape and get an `Arc<Notes>`-holding projection exported under the same name, built through an exhaustively-destructured `From` impl — no `..` — so a field added to the Rust record fails to compile here until the projection is updated too.

Building it is still the same two Nix-then-bindgen steps as before, with the driving script now delegating to the same two scripts CI runs (`nix-build-android-so.sh`, `generate-kotlin-bindings.sh`) instead of a separate copy of that logic, so `just build-kotlin` cannot silently diverge from what `android-native.yaml` / `kotlin-sdk.yaml` do.

The demo app (`android/app`) now has one section per SDK call — wallet status with a live balance, join (preview/join/recover), generate and pay a Lightning invoice, send and redeem ecash, parse an invite code, generate a deposit address, send on-chain, and recent activity — mirroring the layout of `js/examples/vite-core`'s own demo, so every exported call has somewhere to be exercised by hand against a real federation.

## Reviewing

The two places worth being opinionated about, carried over from the original review:

- **Where the Nix/bindgen split falls.** Nix stops at the `.so`; bindings are generated outside it, so each generator is a cheap step over a shared artifact rather than a second derivation re-entering the cross-compile.
- **Retiring the RN lanes rather than disabling them**, since `uniffi-bindgen-react-native` is pinned to uniffi 0.31 and cannot read this crate's 0.32 metadata. The React Native packages under `js/` are untouched.

New in this round:

- **The single-use quote guard (`QuoteClaim`)** sits at the FFI boundary rather than in the plain-Rust `send` methods, since the by-value signature already enforces it there. Worth checking that the three call sites (ecash/lightning/onchain `ffi_send`) all claim before calling into the real `send`, not after.
- **`Notes` as an object vs. a `uniffi::custom_type!` over `String`.** The custom-type form is a few lines shorter but the token would still cross as a plain string, which a generated data class's `toString()` would print. Went with the object for parity with `InviteCode`/`Mnemonic`.
- **The hand-copied `Ffi*` detail records for ecash**, whose only alternative was making the details themselves UniFFI objects (one accessor method per field, and inconsistent with every other facade's plain records).

## Testing

Verified locally on macOS (`nix develop .#android`):

- `cargo fmt --check`, `cargo clippy --locked --tests -- -D warnings` with and without `--features uniffi`, and `cargo doc --no-deps --features uniffi` with warnings as errors — all clean.
- `cargo test --lib` — 529 passing, with and without `--features uniffi`.
- `./scripts/build-android-sdk.sh` — full Nix cross-compile (`arm64-v8a` + `x86_64`) and Kotlin generation.
- `:app:installDebug` onto a Pixel emulator, then exercised by hand against a live mutinynet federation (`fed11qgq...` — the same testnet federation `js/examples/vite-core` pre-fills): open wallet (loads/generates the seed, reattaches to an already-joined federation), preview, join (including the already-joined → reattach path), generate a Lightning invoice, and generate a deposit address, all completed successfully with live-updating operation state.


MrImmortal09:
@fedimint-bot review

rust/fedimint-sdk/src/onchain.rs
```diff
@@ -1625,6 +1700,48 @@ pub(crate) fn deposit_address_of_record(details: &str) -> Option<String> {
         .map(|wire| wire.address)
 }
 
+// The UniFFI views of `Operation<OnchainSendState>` and
+// `Operation<OnchainReceiveState>`: `Operation<S>` is generic and UniFFI
+// objects cannot be, so `crate::ffi::ffi_operation!` monomorphises
+// one newtype object per state, forwarding every method to the real
+// handle. See that macro's documentation in `ffi.rs`.
```
zeenix: nitpick: please use the same line width limit (100 chars) for comments, as the rest of the code. Otherwise we unnecessarily have more LoC than needed. :)

rust/fedimint-sdk/src/sdk.rs
```diff
@@ -1587,13 +1711,43 @@ pub struct FederationInfo {
...
+pub struct FfiFederationInfo {
+    /// See [`FederationInfo::id`].
+    pub id: FederationId,
+    /// See [`FederationInfo::name`].
+    pub name: Option<String>,
+    /// See [`FederationInfo::network`].
+    pub network: Network,
+    /// See [`FederationInfo::status`].
+    pub status: FfiFederationStatus,
+}
```
zeenix: Can we please put all the FFI-specific/only items into a separate module hierarchy?

rust/fedimint-sdk/src/sdk.rs
```diff
@@ -1551,6 +1623,58 @@ pub enum FederationStatus {
...
+        /// [`Diagnostic::message`](crate::Diagnostic::message).
+        message: String,
+    },
+    /// See [`FederationStatus::Closed`].
+    Closed,
+    /// See [`FederationStatus::Forgetting`].
+    Forgetting,
+    /// See [`FederationStatus::Forgotten`].
+    Forgotten,
+}
```
zeenix: Do we really need to duplicate the whole type? 🤔

rust/fedimint-sdk/src/sdk.rs
```diff
@@ -1551,6 +1623,58 @@ pub enum FederationStatus {
...
+#[cfg(feature = "uniffi")]
+impl From<FederationStatus> for FfiFederationStatus {
+    fn from(status: FederationStatus) -> Self {
+        match status {
+            FederationStatus::Running => Self::Running,
+            FederationStatus::Recovering => Self::Recovering,
+            FederationStatus::Quarantined { diagnostic } => Self::Quarantined {
+                code: diagnostic.code,
+                message: diagnostic.message,
+            },
```
zeenix: Review by Codex on behalf of @zeenix.

[P2] Preserve the diagnostic details when exporting quarantine status

This conversion copies only `code` and `message`, discarding `diagnostic.details`. Kotlin
callers of `stored_federations()`, `federation_status()` and status subscriptions consequently
lose structured data such as the conflicting module generations, and unknown detail envelopes
cannot survive the boundary as the contract promises.

Please carry the optional `RawErrorDetails` envelope using `DetailEnvelope::to_raw()`.
The raw-envelope conversion also offers a way to remove these duplicated status/info types;
see the review summary.

android/app/src/main/java/org/fedimint/demo/MainActivity.kt
```diff
@@ -0,0 +1,586 @@
...
+            val lightning = federation?.lightning()
+                ?: return@section "this federation has no lightning module"
+            val msats = amountText.toULongOrNull() ?: return@section "enter an amount in msats"
+
+            val receive = lightning.receive(msats, description)
+            val header = "Invoice:\n${receive.invoice}\n\noperation ${receive.operation.id()}"
+            withContext(Dispatchers.Main) {
+                lastInvoice = receive.invoice
+                lnReceiveCopy.isEnabled = true
+                watch(result, header) { receive.operation.updates().next() }
```
zeenix: Review by Codex on behalf of @zeenix.

[P2] Reuse one subscription for the lifetime of this watcher

The lambda calls `updates()` again on every iteration of `watch`. Each fresh subscription
immediately yields the current state, including a terminal state. The watcher therefore keeps
polling and allocating subscribers instead of waiting for transitions, and it never observes
the final `null` that ends the loop.

Create the subscriber once before starting `watch`, reuse its `next()`, and close it when the
watcher finishes. The recovery, deposit and withdrawal watchers have the same problem.

android/app/src/main/java/org/fedimint/demo/MainActivity.kt
```diff
@@ -0,0 +1,586 @@
...
+        val notesText = findViewById<EditText>(R.id.ecashNotes).text.toString().trim()
+        section(R.id.ecashReceiveResult) {
+            val ecash = federation?.ecash() ?: return@section "this federation has no mint module"
+            if (notesText.isEmpty()) return@section "paste ecash notes first"
+
+            val notes = Notes.parse(notesText)
+            val operation = ecash.receive(notes)
+            val state = operation.awaitFinal()
+            "Redeemed ${formatMsats(notes.value())}\n\n" +
+                "operation ${operation.id()}\nstate: ${describeState(state)}"
```
zeenix: Review by Codex on behalf of @zeenix.

[P2] Only report redemption after a successful terminal state

`awaitFinal()` also returns `EcashReceiveState.Failed` normally: operation failures are
states, while exceptions represent failures to observe the operation. This currently displays
"Redeemed" even when the notes were not redeemed.

Match the returned state and display the success message only for `Done`; display the failure
reason for `Failed`.

zeenix (changes requested):
Review by Codex on behalf of @zeenix. Reviewed `554426a` against T15 and the existing comments.

Requesting changes for the three inline correctness findings. The existing Android build findings
also remain: migrate both modules from `kotlinOptions.jvmTarget` to `compilerOptions`, and align
CI's JDK 17 with the modules' Java 21 target.

Two T15 requirements still need completing:

- Move FFI-specific items into the requested module hierarchy. Keep common operation, subscriber
  and quote adaptation independent of UniFFI so T14 can reuse it. The current `ffi_operation!`
  hardwires UniFFI attributes, and the whole module is enabled only by the `uniffi` feature.
- Add boundary conformance tests. This PR adds no behavioural tests; assembling the demo does not
  check repeated/concurrent quote submission, subscriber completion and cancellation, diagnostic
  preservation, or shutdown waking observers. Exercise those through generated bindings.

On the type/ownership questions, I audited all 23 new Rust types:

| Types | Count | Assessment |
| --- | ---: | --- |
| Concrete operation wrappers | 7 | Justified: UniFFI objects cannot be generic |
| Subscriber wrappers | 7 | Justified: concrete types and serialized mutable access |
| Result records | 4 | Justified: generic operations and nested object references |
| Ecash detail projections | 2 | Justified: preserve opaque Notes and Rust field types |
| Federation status/info projections | 2 | Avoidable; see below |
| QuoteClaim | 1 | Reasonable shared single-use enforcement |

For the duplication comment: export `RawErrorDetails` as a record, map `DetailEnvelope` to it
with `uniffi::custom_type!`, and export the existing `Diagnostic`, `FederationStatus` and
`FederationInfo`. This removes both federation projections while preserving structured details.
I compiled a minimal UniFFI 0.32 prototype and successfully generated Kotlin for this shape.
The complete T15 surface also needs the structured-details accessor on `Error`, which currently
exports only `code()` and `reason()`.

The new `Arc` allocations for objects nested in records, `Option` and `Vec` are required by
UniFFI's conversion traits. As a small simplification, the three `ffi_send` parameters can take
`&Quote`: their bodies only borrow. This does not remove UniFFI's internal shared ownership.
I found no redundant new mutex.

Validation: `cargo check --locked --features uniffi --tests` passed, and I checked the failing
Android CI log. No end-to-end Kotlin tests were run.

rust/fedimint-sdk/src/sdk.rs
MrImmortal09: Yeah sure , moving them rn.

zeenix:
On retiring the React Native lanes: please switch the tool instead of removing them. This PR repoints the Nix cross-compile from `fedimint-client-uniffi` to `rust/fedimint-sdk`'s `uniffi` feature, and the lanes break only because `uniffi-bindgen-react-native` at 0.29.3 cannot read that library's uniffi 0.32 metadata. That pin is lifted in jhugman/uniffi-bindgen-react-native#468 (bump to 0.32). I built its branch head (`Dzejkop:feat/uniffi-bump-to-0.32`, `06f0a30`) and ran it against `libfedimint_sdk.so` built with `--features uniffi`: it reads the metadata and generates the JSI bindings (TypeScript and C++), and the web bindings too. So the React Native packages can bind the same `#[uniffi::export]` surface this PR expands, from the same `.so` the new derivations produce.

Concretely, in this PR rather than a follow-up:

- depend on the tool from that branch, pinned to a commit, in `js/react-native/react-native-bindings/package.json` instead of `0.29.3-1`; its launcher builds the Rust CLI with `cargo run` on first use, and the package's `prepare` script calls `yarn build`, so check what pnpm does with that for a git dependency
- point `ubrn.config.yaml`'s `rust.manifestPath` at `rust/fedimint-sdk/Cargo.toml` with the `uniffi` feature, so the generated JSI bindings cover the new surface; the `--no-cargo` lanes already receive that library from Nix
- keep `.github/actions/build-rn-android`, `.github/actions/build-rn-ios`, the `react-native` change filter and `build-rn-*` jobs in `pull-request.yml` and `snapshot.yml`, and the publish settings on the two `js/react-native/*` packages

T14 (the wasm binding) then builds on this PR: it adds the tool's `web` target and the small crate-side split the feature needs to compile on wasm32 (the tokio runtime pieces native-only, uniffi's `wasm-unstable-single-threaded` on wasm, browser storage in `create_fedimint_sdk`). None of that is needed here.

MrImmortal09:
Regards wasm support , we actually don't know how the exact changes that are required by ubrn for the react-native support ( jhugman/uniffi-bindgen-react-native#468 yet to be merged ), so for the reason I have not included any wasm releated changes here.

MrImmortal09:
@fedimint-bot review

MrImmortal09:
@fedimint-bot review

MrImmortal09:
@zeenix PR is ready for review. 

#353 is going to add more tests to this.

MrImmortal09:
@fedimint-bot review again

js/react-native/react-native-bindings/package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "@fedimint/react-native-bindings",
   "version": "0.1.0",
+  "private": true,
```
zeenix, line 4: Review by Codex on behalf of @zeenix.

Withdrawn: native React Native support and wasm work are deferred to T14 by agreement between
@zeenix and Om. This includes the RN generator/config migration and restoration of its CI and
publishing path. My earlier request relied on the superseded scope and should not block #388.

See [Om's follow-up](#issuecomment-5721503362).

zeenix (changes requested):
Review by Codex on behalf of @zeenix. Rechecked `bf8cc006`, including the newer RN direction.

The three previous correctness findings are fixed: quarantine diagnostics retain their raw detail
envelope, demo watchers reuse and close one subscriber, and failed ecash receives are no longer
labelled redeemed. `Error::details()` is exported, both duplicate federation projections are gone,
and all three send adapters now borrow their quotes. The remaining 21 new types and nested-object
`Arc`/subscriber-lock uses retain concrete purposes; I found no new ownership regression.

The Android compiler DSL/JDK problems are fixed too. Current Rust SDK, Kotlin SDK and Pull Request
CI runs all pass.

Still requesting changes:

- Preserve RN support as requested in the later maintainer comment. The generator/config migration
  is absent, while the RN CI/release lanes and publishing remain removed; see the inline comment.
- Complete the requested FFI module separation. The new hierarchy contains the helper types, but
  FFI-only `ffi_*` adapter impls still live throughout the core modules. I replied on the existing
  thread with the remaining work.
- Add the requested binding conformance coverage. #353 currently contains an Appium mnemonic
  generation test; it does not exercise these new adapters. Cover repeated/concurrent quote use,
  subscription completion/cancellation, detail preservation, and shutdown waking observers through
  generated bindings. Assembling the AAR/demo does not verify those behaviours.

The newer plan to use ubrn's web target in T14 supersedes my earlier request for a separate plain
adapter layer; I am not asking for that alternative architecture or wasm implementation here.

Verification: exact-head source/CI review; no new local Rust or Kotlin test run.

rust/fedimint-sdk/src/sdk.rs
zeenix: Review by Codex on behalf of @zeenix. Rechecked `bf8cc006`: this is only partly addressed.

The helper types/conversions moved, but `ffi/mod.rs` explicitly exempts the FFI-only method blocks.
For example, `Sdk::ffi_federations`, `Federation::ffi_operation`, the `AnyOperation::ffi_as_*`
methods, and each facade's `ffi_send`/`ffi_receive` still live in the core modules.

Please move these adapter impl blocks into the FFI hierarchy as well. Rust permits an inherent
`impl` in another module of the same crate; the derives on the real type definitions can stay.
Adjust helper visibility only where needed. This also fixes comments that currently point at
nonexistent `ffi/federation.rs` or claim adapters live in `ffi/sdk.rs` when they do not.

rust/fedimint-sdk/src/sdk.rs
zeenix: Verified at `bf8cc006` by Codex on behalf of @zeenix: quarantine now exports the real
`Diagnostic`, with its optional raw details preserved. `Error::details()` is also available.

android/app/src/main/java/org/fedimint/demo/MainActivity.kt
zeenix: Verified at `bf8cc006` by Codex on behalf of @zeenix: all four watchers now reuse one
subscriber and close it in `finally`.

android/app/src/main/java/org/fedimint/demo/MainActivity.kt
zeenix: Verified at `bf8cc006` by Codex on behalf of @zeenix: the demo now distinguishes `Done`
from `Failed` and only reports redemption for the successful state.

rust/fedimint-sdk/src/sdk.rs
zeenix: Verified at `bf8cc006` by Codex on behalf of @zeenix: both federation projections are gone.
The existing types are exported directly, with `DetailEnvelope` converted to `RawErrorDetails`.

zeenix (commented):
Review by Codex on behalf of @zeenix.

Scope correction to my previous review: @zeenix has confirmed that the agreement with Om defers
both native React Native support and wasm work to T14. I have withdrawn and resolved the RN
migration/CI/publishing finding. That requirement in my previous summary is superseded and must
not block #388. See [Om's follow-up](#issuecomment-5721503362).

MrImmortal09:
>   Add the requested binding conformance coverage. https://github.com/fedimint/fedimint-sdk/pull/353
>   contains an Appium mnemonic

#353 would cover all the required tests soon.

MrImmortal09:
> * Preserve RN support as requested in the later maintainer comment. The generator/config migration
>   is absent, while the RN CI/release lanes and publishing remain removed; see the inline comment.

This could be added later, right not it is not required. Also generating .so from two different workflows in each PR costs a lot.


> * Complete the requested FFI module separation. The new hierarchy contains the helper types, but
>   FFI-only `ffi_*` adapter impls still live throughout the core modules. I replied on the existing
>   thread with the remaining work.

The impls improve the readability and help developer understand that there are some methods which are common for both uniffi and rust interface, but few which are restricted to only rust and we have uniffi wrappers on them.
This will help in future to make sure no one creates unnecessary wrappers.

MrImmortal09:
@fedimint-bot review and approve if what I am saying is justifiable.

zeenix:
> This will help in future to make sure no one creates unnecessary wrappers.

I don't think that's a problem to solve here.

MrImmortal09:
@fedimint-bot review now

MrImmortal09:
@fedimint-bot review

rust/fedimint-sdk/src/sdk.rs
zeenix: Rechecked `ad87a24e` by Codex on behalf of @zeenix: all 28 FFI-only adapter methods now live
under `ffi`, with their bodies unchanged. The required helper visibility changes are limited
to `pub(crate)`; no new wrapper types or `Arc` layers were added.

The three quote objects retain their UniFFI-gated `QuoteClaim` fields and initializers. Given
@zeenix's latest preference for fewer wrappers, retaining these small fields is a reasonable
exception: moving them out would require separate FFI quote wrappers. The crate-root module,
scaffolding and entry-point re-export gates also remain. Resolving the adapter-separation request.

rust/fedimint-sdk/src/onchain.rs
zeenix: Verified at `ad87a24e` by Codex on behalf of @zeenix: this comment moved to `ffi/onchain.rs`
and now uses the requested wider wrapping. Resolving.

zeenix (approved):
Review by Codex on behalf of @zeenix, rechecked at `ad87a24e`.

All 28 FFI-only adapter methods now live under `ffi`, with their bodies unchanged. The helper
visibility changes are justified, and this update adds no wrapper types or `Arc` layers. The
previous correctness fixes are preserved; I found no new correctness or ownership issues.

The three quote-state fields/initializers remain beside their core types to avoid additional
quote wrappers, consistent with @zeenix's latest preference. I resolved the module-separation
and comment-wrapping threads.

Binding conformance tests are still outstanding and planned for #353. Given the goal of landing
T15 promptly, I am treating that coverage as a follow-up, not claiming it is already delivered.
Please include repeated quote use, subscription completion/cancellation, malformed inputs and
raw diagnostic/error details. Native RN and wasm work remain deferred to T14 as agreed.

Validation: all 28 moved bodies compared against `bf8cc006`; local formatting and diff checks
pass. Current-head CI passes Rust tests (532 unit tests), clippy with and without UniFFI, and
Android native builds, Kotlin generation and AAR/demo assembly.

Approving this revision.

MrImmortal09:
@fedimint-bot create a new issue to improve the ui/ux of the fedimint demo android wallet. ask to add more screens and make it a fully functional wallet.

PR #389, opened 2026-09-11
feat(sdk): implement federation metadata storage and consensus projection facade (T10)

Ayush4958:
### Summary
This PR implements durable storage for federation metadata and aligns the consensus projection logic with the specification.

- **Storage & Recovery:** Adds `ConfigMetaKey` (`0x04`) to persist metadata atomically with federation records, ensuring metadata survives client restarts and is properly restored during recovery.
- **Consensus Projection:** Fixes `apply_consensus_bytes` to strictly adhere to Rule 4, projecting JSON `null` to the literal string `"null"` and correctly merging overlapping consensus/config keys.
- **State Announcements:** Updates `SdkInner::restore` to announce federation status for `Forgetting` and `Closed` states.
- **Integration Testing:** Expands `federation_metadata_persists_across_restart` to actively publish consensus payloads via `fedimint-cli`, asserting config separation, consensus precedence, and proper revision increments.

**Code Generated by Claude Opus 4.6 & Reviewed by Gemini 3.1 Pro**


zeenix (commented):

Reviewed `9c6d0ee` against T10. One numeric-projection correctness issue is noted inline.

T10's required devimint integration coverage is also missing: the added tests cover projection,
database helpers, and closed handles, but never successful metadata reads through a live federation.
Please add coverage for raw-source separation, consensus precedence/revisions, and persistence
across restart/reopen.

The numeric regression was reproduced with the exact projection function and locked serde_json
version; all five existing projection tests passed. CI is green. No other confirmed correctness issues.

rust/fedimint-sdk/src/meta.rs
```diff
@@ -175,6 +202,49 @@ impl Meta {
...
+    let Ok(serde_json::Value::Object(map)) = serde_json::from_str(json_str) else {
+        return merged;
+    };
+    for (key, value) in map {
+        match value {
+            serde_json::Value::String(s) => {
+                merged.insert(key, s);
+            }
+            serde_json::Value::Number(n) => {
+                merged.insert(key, n.to_string());
```
zeenix, line 235: 🤖 ChatGPT (Codex) on behalf of zeenix:

[P2] Preserve numeric values during projection.

With the locked serde_json version/features, `{"n":18446744073709551617}` is projected to
`"1.8446744073709552e+19"`, changing its numeric value. Parsing into `Value` rounds this integer
through `f64` before `to_string()` runs. This violates rule 4's numeric JSON-text projection;
raw consensus bytes remain correct, but `get()`/`all()` report a different value.

Please preserve numeric values without an `f64` round-trip and add a regression test for this case.

zeenix: 🤖 ChatGPT (Codex) on behalf of zeenix:

Confirmed fixed at `bdf53c4`: `arbitrary_precision` preserves the value, and the large-integer regression passes with the locked serde_json version.

rust/fedimint-sdk/src/meta.rs
```diff
@@ -175,6 +202,49 @@ impl Meta {
...
+            }
+            serde_json::Value::Bool(b) => {
+                merged.insert(key, b.to_string());
+            }
+            serde_json::Value::Null => {
+                merged.remove(&key);
```
zeenix: 🤖 ChatGPT (Codex) on behalf of zeenix:

[P2] Restore the specified `null` projection.

Rule 4 still says a JSON `null` contributes the string `"null"`. This new `remove` instead makes `get("k")` return `None` for `{"k":null}` and removes any config value for `k` from `all()`. Both cases should produce `"null"`. The earlier implementation was correct; please restore the insertion and the corresponding test assertion, adding the config-collision case too.

zeenix (commented):

Rechecked `bdf53c4`. The numeric-precision finding is fixed and its thread is resolved. The restore changes now keep federations visible with a quarantine diagnostic on metadata read failure.

One new regression is noted inline: `null` now deletes keys despite rule 4 requiring the string `"null"`. Using the exact projection function and locked serde_json version, the six existing pure projection tests pass; two additional spec tests for `null` (with and without a config value) fail.

The new restart/recovery tests improve coverage, but the consensus part of the earlier request remains incomplete: `federation_metadata_persists_across_restart` only compares snapshots, allowing `None == None`, and never publishes or updates consensus metadata. Please seed an overlapping config/consensus key, assert the raw sources remain separate and `get()`/`all()` prefer consensus, then update it and assert the revision/value change.

CI is green. No other confirmed correctness issues.

rust/fedimint-sdk/src/meta.rs
zeenix: 🤖 ChatGPT (Codex) on behalf of zeenix:

Confirmed fixed at `6f1f0c2`: `null` now yields `"null"` both with and without a config collision. The focused tests pass, including the large-integer regression. Resolving.

zeenix (commented):

Rechecked `6f1f0c2`. The `null` regression is fixed and its thread is resolved. Eight focused tests using the exact projection function and locked serde_json version pass, including both null cases and the large-integer regression. The integration test now exercises published consensus values, precedence, revision changes, and restart behavior.

No remaining production-code correctness findings. One nonblocking issue in the new test's support for the documented four-guardian harness configuration is noted inline; it can be addressed here or tracked in a follow-up. CI is green; the live devimint scenario was inspected rather than executed.

rust/fedimint-sdk/tests/integration.rs
```diff
@@ -1275,3 +1282,146 @@ async fn recovery_locks_the_federation_while_it_runs() {
...
+    let mut words = mint_client.split_whitespace();
+    let program = words
+        .next()
+        .expect("FM_MINT_CLIENT names at least a program");
+    let base_args: Vec<&str> = words.collect();
+
+    let run_admin = |admin_args: &[&str]| {
+        let output = std::process::Command::new(program)
+            .args(&base_args)
+            .args(["--our-id", "0", "--password", "pass"])
```
zeenix: 🤖 ChatGPT (Codex) on behalf of zeenix:

Nonblocking: submit metadata from enough guardians for the configured federation size.

The wrapper explicitly supports `FM_FED_SIZE=4`, but both payloads are submitted only as guardian 0. Upstream records each guardian's proposal separately and changes consensus only after a threshold agrees (three votes for four guardians). Consequently the new test times out waiting for consensus in that supported configuration. Submit each payload for all configured guardian IDs, or at least a quorum. The default single-guardian run is unaffected.

Ayush4958:
@zeenix 
Fixes the all addressed issue

rust/fedimint-sdk/tests/integration.rs
zeenix: 🤖 ChatGPT (Codex) on behalf of zeenix:

Fixed at `5503cba`: both metadata payloads are submitted for every guardian in `0..FM_FED_SIZE`, so four-guardian runs can reach quorum. Resolving.

zeenix (approved):

Rechecked `5503cba`. The guardian-voting fix addresses the remaining finding; all review threads are resolved. Production code and the earlier numeric/null fixes are unchanged. No new issues found. CI is green. The guardian-loop fix was verified by source inspection; I did not rerun live devimint. Ready to merge.

PR #390, opened 2026-09-11
feat(sdk): seed recovery, the lock and the retry (T12)

zeenix:
## Description

- ref fedimint-sdk-implementation-plan T12
- `Sdk::recover` joins a federation and restores this seed's wallet in it; `Sdk::resume_recovery` reattaches to the running attempt, starts a fresh one after a stopped one, or hands back the completed one; `Sdk::recovery_status` reads where a recovery stands without touching the federation. A federation joined with `recover` keeps a record of its current attempt for life, which is how a plainly joined federation (`None`) and a recovered one (`Some(Done)`) stay distinguishable
- the lock is the persisted state, not a flag: a federation is `Recovering` while its rescan has not completed, sends and receives are refused with `Recovering`, a stopped attempt keeps the lock, and a reopen after a stopped attempt mints a new attempt before the client is opened, since the client resumes its rescan by itself the moment it is opened
- the underlying client reports a stopped rescan only to a caller blocked in its wait, and never durably, so one task per recovering federation waits on it: on success it swaps a freshly opened client in under the federation's lock (a v1 mint stays out of the client until the next open after its rescan) and flips the status to `Running`; on failure it records `Failed` with the module's reason. The wait also ends when the federation closes, so a close never finds the client shared
- follows upstream on backups: fedimint has deprecated client backups for removal in 0.13, the mint modules recover from the federation's own recovery log without a snapshot, so `recover` passes none, the docs describe recovery as a rescan, and `Federation::backup` is removed before it was ever implemented
- a `Federation::balance` read during a v1 rescan reports zero, the provisional figure the docs promise, since the mint is absent from the client until the swap; the balance stream wakes on the swap

## Testing

- 413 unit tests (29 new): the record, the wire, the driver's subscription including an ending recorded as it opens, the pre-open rule for every attempt state, the public methods' error paths, the swap refusing a closed federation
- devimint 11/11 on both the v1 and v2 shapes: a wallet funded on one instance is recovered from its seed on a fresh one, spends after recovery (proving the swap made the mint usable), lists the recovery in activity, keeps its state across a restart, and refuses a receive while the lock is observable

### DIFF
--- a/rust/fedimint-sdk/src/federation.rs
+++ b/rust/fedimint-sdk/src/federation.rs
@@ -149,18 +149,27 @@ impl Federation {
         // Reading a balance is not fund-touching: a recovery-locked federation's number is
         // partial and worth showing as progress, and it is the spends that are refused.
         let client = self.inner.client(false).await?;
-        let balance = client
-            .get_balance_for_unit(AmountUnit::BITCOIN)
-            .await
-            .map_err(|err| {
+        let balance = match client.get_balance_for_unit(AmountUnit::BITCOIN).await {
+            Ok(balance) => balance,
+            // A v1 mint recovers as `RecoveryMode::Unusable` and is left out of the client's
+            // module registry until the client is opened again after the rescan
+            // (`fedimint-client/src/client/builder.rs:951`), so a recovering federation of that
+            // generation has no balance source at all yet. What has been recovered *so far* into
+            // something spendable is then exactly nothing, which is the provisional figure the
+            // docs promise; the swap that ends the recovery brings the real one.
+            Err(_) if self.inner.status() == FederationStatus::Recovering => {
+                return Ok(Amount::from_msats(0));
+            }
+            Err(err) => {
                 // The one way this fails in practice is a client with no primary module, which
                 // happens when API-version negotiation left every module out. That is not the
                 // caller's doing and not a storage fault.
-                crate::Error::new(
+                return Err(crate::Error::new(
                     crate::ErrorCode::Internal,
                     format!("this federation cannot report a balance: {err}"),
-                )
-            })?;
+                ));
+            }
+        };
         Ok(Amount::from_msats(balance.msats))
     }
 
@@ -387,20 +396,38 @@ impl BalanceUpdates {
             }
 
             if cursor.stream.is_none() {
+                // Taken before the read below, so a recovery that ends between the read and the
+                // wait on it is not missed.
+                let mut recovery_changed = self.inner.federation.recovery_changed();
                 let client = self.inner.federation.client(false).await?;
                 // Upstream's balance stream never yields and never ends when a client has no
                 // primary module, so a read has to prove there is one before a caller is handed
                 // something that could hang for the life of the process.
-                let initial = client
-                    .get_balance_for_unit(AmountUnit::BITCOIN)
-                    .await
-                    .map_err(|err| {
-                        crate::Error::new(
+                match client.get_balance_for_unit(AmountUnit::BITCOIN).await {
+                    Ok(_) => {}
+                    // No balance source yet because a v1 mint is still recovering (see
+                    // `Federation::balance`): the provisional figure is zero, handed out once,
+                    // and the next change is the recovery ending and swapping a usable client in.
+                    Err(_) if self.inner.federation.status() == FederationStatus::Recovering => {
+                        drop(client);
+                        let zero = Amount::from_msats(0);
+                        if cursor.last != Some(zero) {
+                            cursor.last = Some(zero);
+                            return Ok(zero);
+                        }
+                        tokio::select! {
+                            _ = recovery_changed.changed() => {}
+                            _ = closed.changed() => {}
+                        }
+                        continue;
+                    }
+                    Err(err) => {
+                        return Err(crate::Error::new(
                             crate::ErrorCode::Internal,
                             format!("this federation cannot report a balance: {err}"),
-                        )
-                    })?;
-                let _ = initial;
+                        ));
+                    }
+                }
                 cursor.stream = Some(client.subscribe_balance_changes(AmountUnit::BITCOIN).await);
             }
 
@@ -790,17 +817,7 @@ impl FederationInner {
         &self,
         attempt: fedimint_core::core::OperationId,
     ) -> crate::Result<()> {
-        let record = crate::db::OperationRecord {
-            schema_version: crate::operation::READABLE_STATE_SCHEMA,
-            kind: kinds::RECOVERY.to_owned(),
-            module: String::new(),
-            created_at: crate::db::now_millis(),
-            details: "{}".to_owned(),
-            phase: None,
-            cancel_requested_at: None,
-            final_state: None,
-        };
-        self.write_record(attempt, record, false).await.map(|_| ())
+        record_recovery_attempt_in(&self.db(), attempt).await
     }
 
     /// Looks one operation up by id, rebuilding its record from the client's own log if a crash
@@ -1030,53 +1047,7 @@ impl FederationInner {
         record: crate::db::OperationRecord,
         overwrite_placeholder: bool,
     ) -> crate::Result<crate::db::OperationRecord> {
-        let db = self.db();
-        db.autocommit(
-            |dbtx, _| {
-                let record = record.clone();
-                Box::pin(async move {
-                    let key = crate::db::OperationRecordKey(id);
-                    if let Some(existing) = dbtx.get_value(&key).await {
-                        if !(overwrite_placeholder && is_unclaimed_placeholder(&existing)) {
-                            return Ok::<_, core::convert::Infallible>(existing);
-                        }
-                        // The rebuilt placeholder says nothing the stored one did not already
-                        // say: writing it again would touch storage for no observable
-                        // difference. This is what keeps reconciliation's steady-state pass over
-                        // a module no backfiller has learned to place a pure read, rather than a
-                        // rewrite of the same two rows on every federation open.
-                        if existing == record {
-                            return Ok::<_, core::convert::Infallible>(existing);
-                        }
-                        // Being replaced: its index entry is only still correct if the new
-                        // record keeps the same creation time. A backfill recomputes `created_at`
-                        // from the client's own chronological log, which need not agree with
-                        // whatever an earlier write guessed or was given, and leaving the old
-                        // entry behind would leak it under a key nothing will ever look up again.
-                        if existing.created_at != record.created_at {
-                            dbtx.remove_entry(&crate::db::OperationIndexKey {
-                                created_at: existing.created_at,
-                                id,
-                            })
-                            .await;
-                        }
-                    }
-                    dbtx.insert_entry(&key, &record).await;
-                    dbtx.insert_entry(
-                        &crate::db::OperationIndexKey {
-                            created_at: record.created_at,
-                            id,
-                        },
-                        &(),
-                    )
-                    .await;
-                    Ok::<_, core::convert::Infallible>(record)
-                })
-            },
-            Some(100),
-        )
-        .await
-        .map_err(crate::db::storage_error)
+        write_record_in(&self.db(), id, record, overwrite_placeholder).await
     }
 
     /// A federation handle with no client behind it, for the operation engine's own tests.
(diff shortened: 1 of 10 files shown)


rust/fedimint-sdk/src/recovery.rs
```diff
@@ -222,18 +219,114 @@ impl Sdk {
...
+            Arc::downgrade(self.inner()),
+            self.inner()
+                .db
+                .with_prefix(crate::db::federation_prefix(&id).to_vec()),
+            joined,
+            FederationStatus::Running,
+            Some(client.clone()),
+        ));
+        self.inner().insert(federation.clone());
+        federation.record_recovery_attempt(attempt).await?;
```
zeenix: 🤖 ChatGPT on behalf of zeenix:

**P1 — Set the recovery lock before publishing the federation.**

The federation is inserted as `Running` before this awaited write. A concurrent caller can obtain it through `Sdk::federation`/`federations` and send or receive while the rescan is incomplete, because facade calls do not take the lifecycle mutex. If recording the attempt fails, `recover` returns an error but leaves that live federation permanently `Running`, without starting its watcher. Initialize it as `Recovering` before insertion and preserve the lock on error.

rust/fedimint-sdk/src/recovery/engine.rs
```diff
@@ -0,0 +1,539 @@
...
+    };
+    if client.has_pending_recoveries() {
+        watch(sdk.clone(), federation.clone(), client.clone(), attempt);
+        return FederationStatus::Recovering;
+    }
+    // The client's own durable state already corroborates completion. If the attempt's record
+    // has not caught up yet (the crash window between the client's own commit and this SDK's
+    // write), catch it up now: no watcher starts for an attempt whose client reports nothing
+    // pending, so nothing else ever will.
+    if let Some(record) = read_attempt_record(federation, attempt).await
```
zeenix, line 143: 🤖 ChatGPT on behalf of zeenix:

**P2 — Check module usability when recovery has already finished.**

A v1 rescan can finish between client construction and `after_open`. `has_pending_recoveries()` then returns false, but the mint omitted during construction remains absent until the client is reopened. This branch records `Done` and returns `Running` without the replacement performed by `complete`, so balance reads and spending fail despite successful recovery. Check `all_modules_usable()` and reopen if necessary before reporting readiness; cover a rescan that finishes before this check.

rust/fedimint-sdk/src/federation.rs
```diff
@@ -381,20 +370,38 @@ impl BalanceUpdates {
...
+                        return Err(crate::Error::new(
                             crate::ErrorCode::Internal,
                             format!("this federation cannot report a balance: {err}"),
-                        )
-                    })?;
-                let _ = initial;
+                        ));
+                    }
+                }
                 cursor.stream = Some(client.subscribe_balance_changes(AmountUnit::BITCOIN).await);
```
zeenix, line 412: 🤖 ChatGPT on behalf of zeenix:

**P2 — Recreate established balance streams after client replacement.**

For v2, the mint is usable during recovery, so a subscriber can establish `cursor.stream` before completion. The replacement client creates a new mint balance-notification channel, while this stored stream retains the old module and its channel. Recovery notifications are only handled when `cursor.stream` is absent, so subsequent balance changes never wake this existing subscriber. Invalidate/recreate the stream when the client changes. The integration test should subscribe before completion and keep the same subscription for a post-recovery payment.

rust/fedimint-sdk/src/sdk.rs
```diff
@@ -1826,25 +1884,24 @@ impl SdkInner {
...
+        let outcome: Result<FederationStatus> = async {
+            let attempt = crate::recovery::engine::prepare_open(self, &federation).await?;
+            let (client, revalidated) = self.start(id, &record).await?;
+            // Set before `install`/`announce`, so a refresh `start` made against the client's
+            // live configuration is what `network()`/`capabilities()` and the announced
+            // `FederationInfo` report from here on, not the pre-revalidation snapshot in
+            // `record`.
+            federation.set_record(revalidated);
+            federation.install(client.clone()).await;
+            Ok(crate::recovery::engine::after_open(self, &federation, &client, attempt).await)
```
zeenix: 🤖 ChatGPT on behalf of zeenix:

**P2 — Open the recovery watch state before spawning its watcher.**

`restore` constructs this federation with `Closed`, and `after_open` spawns the watcher before `set_status` below clears its closed flag. On a multithreaded runtime the watcher can run immediately, observe `closed == true`, and exit as `ClientStopped`. The rescan then continues without anything recording its outcome or replacing the v1 client, leaving the SDK stuck in recovery. Establish the open/recovering state before the watcher can run.

rust/fedimint-sdk/src/recovery/engine.rs
```diff
@@ -0,0 +1,539 @@
...
+            target: "fedimint_sdk",
+            federation = %federation.id,
+            attempt = %attempt.fmt_full(),
+            error = %err,
+            "could not record a finished recovery attempt as done",
+        );
+    }
+    federation.bump_recovery();
+    match opened {
+        Ok(()) => federation.set_status(FederationStatus::Running),
```
zeenix, line 265: 🤖 ChatGPT on behalf of zeenix:

**P2 — Serialize completion's status update with close/forget.**

A successful `replace_client` releases the client write lock before the awaited record reads/writes above. `close_federation` can then remove the replacement client and set `Closed`, after which this branch overwrites that status with `Running` and clears the closed watch. The result is a clientless federation advertised as open; forget/shutdown can race similarly. The `FederationClosed` error guard only covers a close that wins before the swap. Serialize the entire completion transition with lifecycle operations, including the final status publication.

rust/fedimint-sdk/src/recovery.rs
```diff
@@ -300,24 +392,94 @@ impl Sdk {
...
+            // record's word: a current attempt whose rescan is not actually live is completed
+            // here, exactly as the watcher would, rather than trusted and handed back to watch.
+            engine::AttemptOnFile::None | engine::AttemptOnFile::Running => {
+                let client = federation.client(false).await?.handle();
+                // Idempotent: repairs the crash window where the record is missing, and is a
+                // harmless rewrite of the same record otherwise. Done before either branch so
+                // that a completion finds a record to mark done.
+                federation.record_recovery_attempt(record.attempt).await?;
+                if !client.has_pending_recoveries() {
+                    engine::complete(self.inner(), &federation, record.attempt).await;
```
zeenix: 🤖 ChatGPT on behalf of zeenix:

**P2 — Drop the cloned client before completing recovery.**

The `client` cloned at line 428 remains alive across this await. `complete` calls `replace_client`, whose shutdown requires the last `Arc<ClientHandle>` reference, so this path necessarily exhausts the approximately two-second reference retry loop. Replacement then proceeds without awaiting the old client's shutdown, while this scope still owns it. Read `has_pending_recoveries`, drop the clone, and only then call `complete`.

zeenix (commented):
🤖 ChatGPT on behalf of zeenix:

Reviewed `0fe8323` against T12. Not ready to merge: six inline findings covering initial recovery locking, fast v1 completion, v2 balance subscriptions, watcher startup, completion versus close/forget, and a retained client handle during resume.

rust/fedimint-sdk/src/recovery.rs
zeenix: Fixed in 252f68a. `recover` builds the federation as `Recovering` before it is inserted, and the attempt record is written before the insert too: on a write failure the client is stopped and the federation is published as `Quarantined` with the error as its diagnostic, which is the outcome the docs already describe for an error after the join has committed. The root record names the attempt, so the reopen writes the missing record under the same id and resumes. `reopen_federation` builds as `Recovering` whenever there is an attempt to resume.

rust/fedimint-sdk/src/recovery/engine.rs
zeenix: Fixed in 252f68a. `after_open` now takes the client by value; when nothing is pending it checks `all_modules_usable`, drops its handle and, if a module is still held back, runs the same swap the watcher would have made before reporting `Running`. Devimint cannot force the rescan to end inside that window, so this path is covered whenever the timing lands there rather than deterministically; the `Done` catch-up remains for the case where every module is usable.

rust/fedimint-sdk/src/federation.rs
zeenix: Fixed in 252f68a. The cursor keeps the `recovery_changed` receiver taken just before the stream was established; the wait selects on it and re-subscribes on the successor client when it fires, and a stream that ends while the federation is still open is treated as a swap rather than a close. `recovery_restores_a_wallet_with_history` now opens one subscription after `recover` returns and keeps it through the swap and the post-recovery payment, on both shapes.

rust/fedimint-sdk/src/sdk.rs
zeenix: Fixed in 252f68a. `after_open` sets the status itself, `Recovering` before it spawns the watcher and `Running` otherwise, so the watch never starts against a federation whose close flag is still set from its construction. `restore` only sets a status of its own on the error path now.

rust/fedimint-sdk/src/recovery/engine.rs
zeenix: Fixed in 252f68a. The watcher completes under the lifecycle mutex: `complete` takes `SdkInner::lifecycle` and runs the swap, the record write and the status publication as one transition, so a `close_federation` or `forget_federation` either finishes before it (the swap is then refused, and the status is left alone) or waits for it. The callers that already hold the mutex call the unlocked `finish` directly. A unit test covers completion against a federation that was closed first.

rust/fedimint-sdk/src/recovery.rs
zeenix: Fixed in 252f68a. `resume_recovery` reads `has_pending_recoveries` through a temporary guard, gone before `finish` runs, and hands its handle to `after_open` by value on the retry path, so the swap finds no stray reference and the shutdown of the retired client completes rather than timing out.

rust/fedimint-sdk/src/recovery.rs
zeenix: 🤖 ChatGPT on behalf of zeenix:

Verified in `252f68a`: recovery is locked before publication, and an attempt-record failure stops and quarantines it.

rust/fedimint-sdk/src/recovery/engine.rs
zeenix: 🤖 ChatGPT on behalf of zeenix:

Verified in `252f68a`: the fast-completion path checks `all_modules_usable()` and drops its handle before replacing an unusable client.

rust/fedimint-sdk/src/federation.rs
zeenix: 🤖 ChatGPT on behalf of zeenix:

Verified in `252f68a`: established balance streams observe recovery changes and reattach after replacement. A separate comment covers a timing bug in the updated integration test.

rust/fedimint-sdk/src/sdk.rs
zeenix: 🤖 ChatGPT on behalf of zeenix:

Verified in `252f68a`: `after_open` sets `Recovering` before spawning the watcher.

rust/fedimint-sdk/src/recovery/engine.rs
zeenix: 🤖 ChatGPT on behalf of zeenix:

Verified in `252f68a`: watcher completion holds the lifecycle mutex across replacement, persistence and status publication. A separate comment covers duplicate completion of the same attempt.

rust/fedimint-sdk/src/recovery.rs
zeenix: 🤖 ChatGPT on behalf of zeenix:

Verified in `252f68a`: the temporary client guard is dropped before `finish`, and `after_open` receives the retry handle by value.

rust/fedimint-sdk/src/recovery/engine.rs
```diff
@@ -0,0 +1,605 @@
...
+/// `close_federation` or `forget_federation` interleaved between the two would have its
+/// `Closed` overwritten with `Running` on a federation whose client had just been taken away.
+/// The mutex is taken before anything else, as every lifecycle call takes it.
+pub(crate) async fn complete(
+    sdk: &Arc<SdkInner>,
+    federation: &Arc<FederationInner>,
+    attempt: UpstreamOperationId,
+) {
+    let _lifecycle = sdk.lifecycle.lock().await;
+    finish(sdk, federation, attempt).await;
```
zeenix, line 230: 🤖 ChatGPT on behalf of zeenix:

**P2 — Recheck completion after acquiring the lifecycle mutex.**

`resume_recovery` can hold this mutex while the upstream rescan ends. It sees the attempt record still `Running` and `has_pending_recoveries() == false`, calls `finish`, installs a usable client and records `Done`. The watcher meanwhile drops its old handle and queues here. Once resume releases the mutex, the watcher unconditionally calls `finish` again, shutting down the now-live replacement and opening it a second time. Fund-touching calls are already permitted after the first completion, and a failure of this unnecessary reopen quarantines an already recovered federation. Check the attempt's current state under the mutex and skip the swap if it has already completed. Cover resume winning the mutex before the watcher.

rust/fedimint-sdk/tests/integration.rs
```diff
@@ -954,3 +981,293 @@ async fn activity_lists_what_the_federation_was_used_for() {
...
+    assert_eq!(last, RecoveryState::Done);
+    assert_eq!(
+        sdk_b.recovery_status(&id).await.expect("readable"),
+        Some(RecoveryState::Done)
+    );
+    assert_eq!(
+        sdk_b.federation_status(&id),
+        Some(FederationStatus::Running)
+    );
+    settles_at(&mut updates, funded).await;
```
zeenix: 🤖 ChatGPT on behalf of zeenix:

**P2 — Account for the initial balance already equalling `funded`.**

The first `updates.next()` above can already return `funded` when recovery finishes quickly; the preceding assertion explicitly permits equality. This call then waits for that same value again on the same subscription. `BalanceUpdates` deduplicates against its last returned value even across client replacement, and the test does not initiate its payment until after this wait, so it times out after 60 seconds despite correct recovery. Skip `settles_at` when `provisional == funded`, or pass the already-consumed value into the helper.

zeenix (commented):
🤖 ChatGPT on behalf of zeenix:

Rechecked `252f68a`. The original six findings are addressed and their threads are resolved. Two remaining P2 findings below: completion needs to be idempotent across resume and the watcher, and the updated integration test can wait for a balance it already consumed. CI is green.

rust/fedimint-sdk/src/recovery/engine.rs
zeenix: Fixed in 35fbce5. `complete` re-reads the attempt's record once it holds the lifecycle mutex and returns without swapping when a final state is already recorded, so a completion `resume_recovery` made while holding the mutex is not repeated by the queued watcher. A unit test plants a concluded attempt and checks a second `complete` neither rewrites it nor wakes its subscribers.

rust/fedimint-sdk/tests/integration.rs
zeenix: Fixed in 35fbce5. The wait for the funded figure is skipped when the provisional reading already was that figure, since the stream reports changes only and would never yield it again. The subscription is still the one that reports the payment afterwards.

rust/fedimint-sdk/src/recovery/engine.rs
zeenix: 🤖 ChatGPT on behalf of zeenix:

Verified in `35fbce5`: the watcher re-reads the attempt under the lifecycle mutex and skips an already-concluded attempt. The new regression test covers that no-op path. Resolved.

rust/fedimint-sdk/tests/integration.rs
zeenix: 🤖 ChatGPT on behalf of zeenix:

Verified in `35fbce5`: the test skips the repeated wait when the initial reading already equals `funded`, while retaining the subscription for the later payment. Resolved.

zeenix (commented):
🤖 ChatGPT on behalf of zeenix:

Rechecked `35fbce5`: both remaining findings are fixed, and all eight review threads are resolved. The rebase preserves the earlier recovery fixes. SDK tests, clippy, fmt, docs, wasm and pin checks pass in CI. No further blockers found; ready to merge from this review.

MrImmortal09 (approved):
LGTM.

Issue #370, opened 2026-09-02
Finish outcome-neutral wording for transfer terms


zeenix:
🤖 ChatGPT GPT-5.6 on behalf of @zeenix:

PR #369 establishes the right general contract: persisted details and `ActivityItem` describe the authorized or attempted terms for every outcome, and only a successful terminal state turns those terms into realized balance movement. A few rustdoc remnants still state unsuccessful operations as completed facts:

- `activity.rs:52-53` says the fee “was charged,” while the field contract at lines 182-195 correctly says it is what the wallet pays if the transfer succeeds.
- `activity.rs:85-91` says every `Refunded` or `Canceled` transfer went out and came back. Funding can be rejected before a lasting debit, and an expired/canceled incoming Lightning invoice was never paid.
- `lightning.rs:592-594` calls every persisted send record an invoice “that was paid,” even though the next bullet correctly requires the same record for refunded and failed sends.

Please make these passages outcome-neutral without changing the public API:

- terms describe what was attempted/authorized;
- `Success` says those terms were realized;
- `Refunded` says there is no lasting debit and the funds are safe, without prescribing whether value never left or was restored;
- `Canceled` covers both reclaimed outgoing value and incoming transfers that never happened.

This is non-blocking follow-up from #369.

sadeeqrabiu:
I will also work on this too

zeenix:
@sadeeqrabiu didn't #391 fixed this? Please remember to put `Fixes #<number>.` line in the PR description for issues to be closed by PR merges. 🙏

zeenix:
Fixed by #391.

Issue #371, opened 2026-09-02
Emit send Refunded only after rejected funding is spendable


zeenix:
🤖 ChatGPT GPT-5.6 on behalf of @zeenix:

The app-facing `Refunded` contract introduced in #369 is the right one: it is final and tells an app that the payment/withdrawal did not complete and the funds are again spendable. The implementation must not emit it merely because an upstream Lightning or wallet send subscription reports that its funding transaction was rejected.

Submitting a client transaction removes/selects primary-module inputs and their recovery can continue asynchronously after rejection. Therefore the SDK needs to keep the operation non-final until the relevant input recovery has settled, then emit:

- `Refunded` only once the authorized value is known to be spendable again;
- `Failed` if the monetary effect cannot be established as a clean return.

The rustdoc should also stop describing rejected funding as “nothing was debited” / “never debited.” It can promise no lasting debit once `Refunded` is emitted. Likewise, exact-quote wording such as “debits this or does not run” should distinguish exact authorized debit from a rejected funding attempt that may temporarily remove inputs and later restore them.

Cover both Lightning and on-chain send paths and restart/reattachment in implementation tests. Current upstream recovery/quiescence gaps are tracked by [fedimint#9099](https://github.com/fedimint/fedimint/issues/9099), [#6546](https://github.com/fedimint/fedimint/issues/6546), and [#8421](https://github.com/fedimint/fedimint/issues/8421). Keep the SDK contract strong; if Fedimint 0.12 cannot prove restoration yet, ignore the blocked implementation test with the relevant upstream link rather than weakening the API.

This is non-blocking follow-up from #369.

sadeeqrabiu:
I will work on this

zeenix:
@sadeeqrabiu I believe this was also fixed by #391?

sadeeqrabiu:
Yeah @zeenix

zeenix:
Fixed by [#391](https://github.com/fedimint/fedimint-sdk/pull/391).

PR #391, opened 2026-09-14
fix(sdk): gate send Refunded on input recovery settlement

sadeeqrabiu:
- Settle primary-module input recovery via inputs::settle before emitting
  terminal Refunded for rejected funding transactions on Lightning sends
- Emit LnSendState::Failed if input recovery cannot be proven clean
- Introduce SendStep in lightning::driver to decouple upstream rejection from
  terminal state emission
- Update activity, lightning, and onchain rustdoc passages to be outcome-neutral
  regarding fees, attempted terms, and lasting debits
- Add unit tests for restoration verdict logic and gated stream transitions
- Add integration tests for Lightning and onchain rejected funding paths
  (ignored pending upstream fedimint#6546, #9099, #8421)
Addresses #370, #371


rust/fedimint-sdk/src/lightning/driver.rs
```diff
@@ -148,7 +199,7 @@ fn unknown_module(module: &str) -> Error {
 mod tests {
     use std::time::Duration;
 
-    use futures::{StreamExt as _, stream};
+    use futures::stream;
```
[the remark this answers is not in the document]
sadeeqrabiu: No change needed here — in Rust, `use super::*;` brings the parent's anonymous trait imports (`use futures::StreamExt as _;`) into the child `mod tests` scope, so `.next()` is already available and compiles cleanly.

Re-adding a local `StreamExt` import in `mod tests` causes `cargo clippy --all-targets -- -D warnings` to fail with `error: unused import: StreamExt`.

rust/fedimint-sdk/src/inputs.rs
```diff
@@ -0,0 +1,229 @@
...
+    let mut unresolved = 0usize;
+    for state in submissions {
+        match state {
+            TxSubmissionStates::Accepted(_) => accepted += 1,
+            TxSubmissionStates::Rejected(..) => rejected += 1,
+            // `Created` after the operation has gone quiet is a transaction whose submission
+            // never resolved either way, and `NonRetryableError` is one that gave up. Neither
+            // says where the value is.
+            TxSubmissionStates::Created(_) | TxSubmissionStates::NonRetryableError(_) => {
+                unresolved += 1;
```
zeenix: 🤖 ChatGPT (Codex) on behalf of zeenix:

[P1] Historical `Created` states make successful recoveries report `Failed`

`get_operation_states` includes the entire inactive history: the executor retains the old `Created` state when inserting `Accepted` or `Rejected`. A completed funding-rejection/recovery sequence therefore contains `Created(funding), Rejected(funding), Created(recovery), Accepted(recovery)`. This code counts two unresolved transactions and returns `Unproven` even when recovery succeeded. Match historical states by transaction ID and distinguish superseded `Created` entries from genuinely unresolved submissions; add a regression using that history.

rust/fedimint-sdk/src/inputs.rs
```diff
@@ -0,0 +1,229 @@
...
+                .to_owned(),
+        );
+    }
+    if accepted == 0 {
+        return Restoration::Unproven(
+            "this payment's funding was rejected and nothing the SDK can see put the value back"
+                .to_owned(),
+        );
+    }
+    Restoration::Restored
```
zeenix: 🤖 ChatGPT (Codex) on behalf of zeenix:

[P1] Transaction acceptance does not establish successful note issuance

Fixing the historical-`Created` problem is necessary, but this verdict would still accept insufficient evidence: an accepted recovery transaction can have mint outputs that terminate in `Failed`/`Failure` without issuing spendable notes. `has_active_states == false` also holds in that case, and this code ignores those output outcomes. Verify the recovery outputs succeeded before returning `Restored`; if the pinned upstream API cannot establish that, preserve `Unproven` rather than asserting the funds are spendable.

rust/fedimint-sdk/src/lightning/driver.rs
```diff
@@ -1,18 +1,69 @@
...
+) -> BoxStream<'static, Result<LnSendState>> {
+    Box::pin(stream.then(move |step| {
+        let sdk = sdk.clone();
+        async move {
+            match step? {
+                SendStep::State(state) => Ok(state),
+                SendStep::FundingRejected => {
+                    match crate::inputs::settle(sdk, federation_id, id).await? {
+                        Restoration::Restored => Ok(LnSendState::Refunded),
+                        Restoration::Unproven(reason) => Ok(LnSendState::Failed { reason }),
```
zeenix: 🤖 ChatGPT (Codex) on behalf of zeenix:

[P2] Yield a pending state before waiting on a cached rejection

Upstream `outcome_or_updates` can return only a cached terminal funding-rejection outcome. When reattaching to such an operation while input recovery is still running, this becomes the first step, and `then` waits for settlement before yielding anything. `settled` awaits its first item without a timeout, so `Operation::state()` and a new subscription's first update can wait indefinitely instead of reporting the current pending state. Ensure this path yields a non-final state before the recovery wait, and test a stream containing only `FundingRejected`.

rust/fedimint-sdk/src/lightning.rs
```diff
@@ -300,12 +300,25 @@ impl LnQuote {
...
+    ///
+    /// "This much or nothing" is a statement about the authorised debit, not
+    /// about what the balance does moment to moment. Executing the quote
+    /// submits a transaction, and submitting one takes the notes that are to
+    /// pay for it out of the spendable set before the federation has
+    /// accepted anything. A funding attempt the federation then rejects can
+    /// therefore remove value and restore it afterwards. Nothing else is
+    /// ever debited than this total, and a payment that does not succeed
+    /// leaves no lasting debit at all, but the balance may dip and recover
+    /// in between; [`LnSendState::Refunded`] is the state that says the
```
zeenix: 🤖 ChatGPT (Codex) on behalf of zeenix:

[P2] Restrict the no-lasting-debit guarantee to `Refunded`

A payment that does not succeed may end in `Failed` because its inputs or contract could not be recovered. Consequently, the unconditional statement that every unsuccessful payment leaves no lasting debit contradicts the failure-state contract and the new gate's own `Unproven` outcome. Promise restoration only when `Refunded` is reached. Please fix the equivalent wording in `OnchainQuote::total` (`onchain.rs:306–307`) too.

zeenix (changes requested):

Reviewed `d0169b3`. Requesting changes for the recovery-state handling and the unconditional restoration guarantee in the quote documentation; details inline.

Isolated reproductions using the PR's functions confirm the historical-state misclassification and the missing initial state for a cached rejection. Please cover actual executor history, failed output issuance, and cached-outcome reattachment. The new integration tests are ignored and their `rejected_funding` helper is still `unimplemented!()`.

rust/fedimint-sdk/src/inputs.rs
sadeeqrabiu: Updated in commit `3db16af`.
Instead of raw downcast counting over the flat state list, we now group history entries by `TransactionId` via `submissions_of`:
1. Each unique submission tracks its terminal state (`Accepted`, `Rejected`, or still pending/hanging `Submitted`).
2. Superseded `Created` states retained in the executor's inactive history for a transaction that subsequently reached `Accepted` or `Rejected` are treated as resolved and no longer counted as hanging.
3. Only transactions whose final observed state is non-terminal or non-retryable error are flagged as unresolved.
Added regression unit tests:
- `a_superseded_created_is_not_an_unresolved_transaction` (asserting `[funding: Submitted, funding: Rejected, recovery: Submitted, recovery: Accepted]` settles cleanly to `Evidence::Recovery`)
- `the_history_may_arrive_in_any_order` (verifying order-independence across inactive state fetches)

rust/fedimint-sdk/src/inputs.rs
sadeeqrabiu: `Refunded` must strictly guarantee spendable notes, not merely transaction consensus acceptance.

In commit `3db16af`, after transaction acceptance is established, the settle engine explicitly queries `client.await_primary_bitcoin_module_output(id, out_point)` for each recovery output.
- If any output reports an error or terminal failure to finalize and issue its ecash notes, it returns `Restoration::Unproven("...output X of TX did not issue its notes: ...")`.
- If an output does not report an outcome within a bounded timeout (5s), it also returns `Restoration::Unproven` rather than guessing.
- `Restoration::Restored` is emitted only when every recovery output has been verified to have produced spendable notes.
Added unit tests:
- `a_recovery_output_that_failed_to_issue_is_unproven`
- `an_output_that_never_reported_is_unproven`
- `a_recovery_whose_outputs_all_issued_is_a_restoration`

rust/fedimint-sdk/src/lightning/driver.rs
sadeeqrabiu: Fixed in commit `3db16af`. 
`through_settle` was refactored using `futures::stream::unfold` into a 2-stage state machine (without requiring `Send`, preserving `wasm32` compatibility):
1. Upon encountering `SendStep::FundingRejected`, the stream immediately yields `Ok(LnSendState::Created)` to callers so `Operation::state()` and new subscriptions receive the current pending state without blocking.
2. It then arms `settle_next = true`, awaiting `crate::inputs::settle` on the subsequent pull before yielding the final ending (`Refunded` or `Failed`).
Added unit test:
- `a_cached_rejection_reports_pending_before_waiting` (verifies that a stream containing only `SendStep::FundingRejected` yields `LnSendState::Created` as its first non-final item before settling).

rust/fedimint-sdk/src/lightning.rs
sadeeqrabiu: Both [`LnQuote::total`] and [`OnchainQuote::total`] rustdoc passages have been aligned:
- The unconditional statement has been removed.
- They now explicitly state that the quote specifies only the authorized maximum debit, and that what becomes of value on failure is the ending's contract: `Refunded` is the state that guarantees no lasting debit once spendable, whereas `Failed` cannot promise it.

rust/fedimint-sdk/tests/integration.rs
```diff
@@ -1433,3 +1433,249 @@ async fn federation_metadata_persists_across_restart() {
...
+/// un-ignoring them is the whole change once upstream lands.
+async fn rejected_funding(
+    _devimint: &Devimint,
+    _sdk: &Sdk,
+    _federation: &fedimint_sdk::Federation,
+    _lightning: &fedimint_sdk::Lightning,
+) -> Option<(Sdk, fedimint_sdk::Operation<fedimint_sdk::LnSendState>)> {
+    // fedimint/fedimint#6546: restore a second client onto the same notes, spend them there,
+    // then quote and send here so this send's funding transaction is refused.
+    None
```
zeenix: 🤖 ChatGPT (Codex) on behalf of zeenix:

[P2] Keep the blocked fixture from silently passing when enabled

Changing the helper from `unimplemented!()` to unconditional `None`, together with the new caller early returns, makes both Lightning rejection tests report success without reaching any rejection/recovery assertions—even with a federation available and `--ignored`, or after removing `#[ignore]`. The comments saying un-ignoring is sufficient are therefore incorrect. Keeping these tests `#[ignore]` with the upstream links is fine; leave the missing fixture explicitly failing when deliberately invoked (or implement it), and describe the remaining work accurately.

js/.changeset/gated-send-refunded.md
```diff
@@ -0,0 +1,19 @@
+---
+'@fedimint/core': major
```
zeenix: 🤖 ChatGPT (Codex) on behalf of zeenix:

[P2] Remove the unrelated major bump for `@fedimint/core`

This changeset queues a major release of the published JavaScript package, but this PR only changes the new `rust/fedimint-sdk` implementation. `@fedimint/core` still uses the existing RPC client (`LightningService.subscribeLnPay` calls `subscribe_ln_pay`); its web/native backends remain the upstream WASM bundle and `fedimint-client-uniffi`. None consumes this new `LnSendState` gate yet. The changelog would therefore announce behavior JavaScript users do not receive, while triggering a major package bump. Remove this npm changeset and document the change with the Rust SDK until those bindings actually adopt it.

zeenix (changes requested):

Rechecked `3db16af`: all four original findings are addressed. Transaction histories now reconcile by ID, every recovery output is checked through the primary-module API, cached rejections yield a pending state before settlement, and both quote docs distinguish `Refunded` from `Failed`.

CI is green. Sixteen isolated checks using the current history/verdict/stream functions passed, including cancellation during settlement and cached-outcome reattachment. Two follow-up changes still need correcting: the unconditional skip in the blocked tests and the unrelated JavaScript major-version changeset. Details inline.

rust/fedimint-sdk/src/onchain.rs
```diff
@@ -530,14 +542,25 @@ pub enum OnchainSendState {
...
+    ///
+    /// The two halves of that are separate events and this state waits for
+    /// the second. Funding selects the notes that are to pay for the
+    /// withdrawal before consensus has accepted anything, so the rejection
+    /// leaves value that is neither spent nor yet spendable, and returning
+    /// it is a later transaction of the mint's own. The withdrawal stays
+    /// non-final while that runs, and reports this state only once it has
+    /// settled in a way that establishes the value came back. A recovery
+    /// that settles without establishing it ends in
+    /// [`Failed`](Self::Failed) instead.
```
zeenix, line 563: 🤖 ChatGPT (Codex) on behalf of zeenix:

[P1] Connect the rebased on-chain drivers to the recovery gate

After rebasing onto #392, this is an executable path, but the guarantee added here is not implemented. `onchain/v1.rs::map_withdraw` maps `WithdrawState::Failed` directly to `Refunded`; `onchain/v2.rs::map_final_send` does the same for `Aborted`, and both `current_send` and `subscribe_send` use it. None calls `inputs::settle`. A rejected withdrawal can therefore be reported and persisted as final `Refunded` while its input recovery is still running, or even when recovery later fails. The previous implementation note requiring this gate disappeared in the rebase. Route both wallet generations through settlement, including walletv2's separate current-state path, and cover pending recovery, failed recovery, and cached-outcome reattachment as required by #371.

zeenix (changes requested):

Rechecked `0bea4ec`, including the rebase onto #392. The original four fixes remain intact, the blocked Lightning fixture now fails explicitly when deliberately invoked, and both CI workflows are green.

The rebase brings in executable on-chain send drivers, but the recovery gate has not been connected to them: the new on-chain guarantee is still violated on funding rejection. Details inline. The previous npm changeset finding is only partially addressed: reducing the bump to `patch` still releases and documents a change that `@fedimint/core` does not consume.

js/.changeset/gated-send-refunded.md
zeenix: 🤖 ChatGPT (Codex) on behalf of zeenix:

Rechecked `0bea4ec`: the bump is now `patch`, but this file still targets `@fedimint/core` and announces recovery behavior that package does not receive. The release workflow explicitly runs Changesets with `cwd: js`, so this remains a package release entry. Please remove this npm changeset; changing its release type only addresses the major-bump portion of the finding. Reopening until the unrelated entry is removed.

rust/fedimint-sdk/src/onchain.rs
sadeeqrabiu: Following the rebase onto #392, the recovery settlement gating engine has been connected to both wallet module generations:
### 1. Generalized Settle Gate (`inputs.rs`)
- Extracted the settlement stream state machine into `crate::inputs::through_settle<S, R, F>`:
  - Takes arbitrary state enum `S` (`LnSendState` or `OnchainSendState`).
  - Carries module rejection diagnostic `R` (`()` for Lightning, `String` reason for Onchain).
  - Emits non-final `pending` (`Created`) immediately upon encountering `FundingRejected` to prevent stalling `Operation::state()` and new subscriptions on cached rejections.
  - Passes through `inputs::settle` and maps `Restoration::Restored` $\rightarrow$ `Refunded` and `Restoration::Unproven` $\rightarrow$ `Failed { reason }`.
### 2. On-Chain Send Driver (`onchain/driver.rs`)
- Introduced `SendStep`:
  ```rust
  pub(super) enum SendStep {
      State(OnchainSendState),
      FundingRejected { reason: String },
  }

rust/fedimint-sdk/src/onchain/v2.rs
```diff
@@ -287,12 +300,24 @@ pub(super) async fn current_send(
...
 
+/// What [`current_send`] reports for a final state upstream has already recorded.
+///
+/// Split out from the client call so the one decision it makes can be tested on its own: a
+/// funding rejection is not an ending, and this path cannot run the gate that turns it into
+/// one, so it reports the withdrawal as still running.
+fn current_of(state: &FinalSendOperationState) -> OnchainSendState {
+    match map_final_send(state) {
+        SendStep::State(state) => state,
+        SendStep::FundingRejected { .. } => OnchainSendState::Created,
```
zeenix: 🤖 ChatGPT (Codex) on behalf of zeenix:

[P2] Let current-state reads observe completed recovery

This maps every cached `Aborted` to `Created`, even after recovery has fully succeeded or failed. With no persisted SDK final state, `OnchainSendDriver::current` always takes this path, so an application polling only `Operation::state()` stays pending forever, including after restart. Upstream's cached outcome remains `Aborted`; only consuming a separate subscription can persist the SDK ending and make polling see it. The bounded-read requirement does not require discarding completed recovery: use the same settled-subscription/first-state path as wallet v1, which reports pending after the bounded wait but returns an available terminal result. Add coverage for repeated state reads after both successful and failed recovery without an existing subscription.

zeenix (changes requested):

Rechecked `be6bd67`. Both wallet subscriptions now gate funding rejection on recovery, and the shared-stream refactor preserves the Lightning fixes. CI is green. Nineteen isolated checks passed, covering history reconciliation, output verdicts, pending recovery, cancellation, and both send-state wrappers. One additional check reproduces the walletv2 current-state regression below: completed recovery yields `Refunded` through a fresh subscription but `Created` through the current-state mapping.

Please fix that polling path. The existing npm changeset finding also remains unaddressed; the unrelated `@fedimint/core` patch entry is unchanged.

zeenix (approved):

Rechecked `2fa8f89`: both remaining findings are fixed. Walletv2 current-state reads now use the settled subscription path and observe completed recovery; the unrelated npm changeset is removed. Earlier recovery fixes remain intact, and I found no new issues in this update.

CI is green. All 21 isolated checks pass, including repeated polling from pending to successful or failed recovery without an existing subscription, cached outcomes, and cancellation. The upstream-blocked integration fixtures remain explicitly ignored. Resolved the two remaining review threads.

sadeeqrabiu:
@zeenix This PR Fix #370 and #371

PR #392, opened 2026-09-15
feat(sdk): the on-chain facade over either wallet generation (T9)

zeenix:
## Description

- ref fedimint-sdk-implementation-plan T9
- `Onchain::receive` hands out a fresh deposit address with one durable operation behind it, `Onchain::quote` plans a withdrawal and freezes its destination, amount, aggregate fee and total, and `Onchain::send` executes exactly that or refuses with `QuoteChanged`; both wallet module generations sit behind the one surface
- on walletv2 an address has no operation behind it upstream: the module creates one, under a random id, only when its background scanner claims a deposit. The SDK mints the operation id itself at `receive` time and the driver finds the claim afterwards through the client's event log, matching on the address; the reconciler skips an upstream walletv2 receive entry whose address an SDK record already owns, so one deposit is one row
- a deposit's fee and net credit are read from the claim transaction itself (its mint outputs and the mints' own fee schedules), not from a fee dry-run: the mint shapes its change to a target note count per denomination, so a dry-run made after the claim sees a different inventory and a different split. That read goes through `Client::executor`, which upstream marks as internal tooling; the follow-up is an upstream accessor for what a claim minted
- a v1 deposit at or below the flat deposit fee maps to `Failed`: upstream reports it claimed while minting nothing
- an unfunded address never blocks `forget_federation`; a deposit that has been seen but not claimed does, through a phase the driver records
- v1 `Failed` and walletv2 `Aborted` withdrawals map to `Refunded`; walletv2 `Failure` is the one `Failed`

## Testing

- 487 unit tests (37 new, 495 with the T10 merge): the wire records, the quote's checks and fee arithmetic, both generations' state mappings and backfills, the claim transaction's shape check, the drivers, the adoption lookup and the erase guard
- devimint 15/15 on both the v1 and the walletv2 shapes: a deposit paid through `bitcoin-cli` is seen, claimed and credited with the exact net figure the record promises, lands in activity, and a withdrawal quoted and sent from it reaches the destination for the quoted amount with the quoted total debited; a deposit address survives a restart and is paid afterwards under the same operation id; the quote refuses a foreign-network address, zero, dust and an amount above the balance


rust/fedimint-sdk/src/onchain.rs
```diff
@@ -853,14 +909,544 @@ struct OnchainInner {
...
+/// from the wallet module's own instance (a claim funds itself from one wallet input, and
+/// nothing else), and at least one output to read credit from. Split out of
+/// [`claim_mint_outputs`] so it is testable without a client or a real transaction.
+fn check_claim_shape(
+    inputs: usize,
+    input_instance: Option<ModuleInstanceId>,
+    wallet_instance: ModuleInstanceId,
+    outputs: usize,
+) -> Result<()> {
+    if inputs != 1 {
```
zeenix: 🤖 ChatGPT (Codex) on behalf of zeenix:

[P1] Accept claim transactions that rebalance existing ecash

A valid claim can have mint inputs as well as its wallet input. At this pin, mint v1's `create_final_inputs_and_outputs` always calls `consolidate_notes` (triggered above eight notes in a denomination), and mint v2's `rebalance` can also consume existing notes. This check therefore makes a successfully credited deposit return `Internal` forever instead of `Claimed` on an ordinary funded wallet. The empty-wallet integration test misses it.

Read the actual net mint movement (mint outputs minus pre-existing mint inputs), including input fees, rather than requiring exactly one input or merely removing the check. Please cover a deposit into an inventory that triggers consolidation/rebalancing.

rust/fedimint-sdk/src/onchain/v2.rs
```diff
@@ -0,0 +1,1202 @@
...
+                            ReceiveCursor::Linking {
+                                address,
+                                cursor: from,
+                                announced: true,
+                            },
+                        ));
+                    }
+                    None => {
+                        let mut closed = federation.closed();
+                        tokio::select! {
```
zeenix, line 975: 🤖 ChatGPT (Codex) on behalf of zeenix:

[P1] Release the client handle before waiting for another event

The `handle` cloned above remains in scope across this `select!`. After receiving `WaitingForTransaction`, time out a subsequent `updates.next()` and retain the subscriber: its stored unfold future keeps that handle, and neither branch of this select is polled again. `close_federation`/`shutdown` then fail when `shutdown_client` cannot unwrap the client after its two-second retry window.

Drop the handle before the event wait. Also move the client-owning async reads in this stream onto `wait_holding_client`, as its documentation requires; a cancelled poll can otherwise park those reads too. Add a regression that retains the subscriber after cancelling `next()` and then closes the federation.

rust/fedimint-sdk/src/onchain/v1.rs
```diff
@@ -0,0 +1,799 @@
...
+/// for a withdrawal, since neither is knowable from the operation log alone.
+pub(super) fn backfill(
+    id: OperationId,
+    meta: &serde_json::Value,
+    created_at: u64,
+) -> Option<Backfilled> {
+    let meta: WalletOperationMeta = serde_json::from_value(meta.clone()).ok()?;
+    match meta.variant {
+        WalletOperationMetaVariant::Deposit { address, .. } => {
+            let wire = from_custom_meta::<wire::OnchainReceiveDetailsWire>(&meta.extra_meta)
```
zeenix: 🤖 ChatGPT (Codex) on behalf of zeenix:

[P2] Replace the deposit address placeholder during backfill

`receive()` deliberately stores `address: String::new()` in upstream custom metadata. If the process stops after address allocation commits but before `create_operation`, this branch successfully decodes that placeholder and never uses the real address from `WalletOperationMetaVariant::Deposit`. The reconstructed record consequently fails `decode_receive_details`, leaving the recovered operation unreadable. Always fill the address from upstream metadata, including when the custom metadata exists; test that exact crash-window payload.

rust/fedimint-sdk/src/onchain/v2.rs
```diff
@@ -0,0 +1,1202 @@
...
+) -> Result<OnchainReceiveState> {
+    let entry = client
+        .operation_log()
+        .get_operation(found.upstream)
+        .await
+        .ok_or_else(|| internal("the linked deposit is no longer in the operation log"))?;
+    let meta: WalletOperationMeta = entry
+        .try_meta()
+        .map_err(|err| internal(format!("could not read the linked deposit's terms: {err}")))?;
+    let WalletOperationMeta::Receive(ReceiveMeta { value, fee, .. }) = meta else {
```
zeenix: 🤖 ChatGPT (Codex) on behalf of zeenix:

[P2] Wait for mint issuance before recording Claimed

Walletv2's `FinalReceiveOperationState::Success` means the claim transaction was accepted, not that its mint outputs are spendable. This function reads the submitted transaction and permanently records `Claimed` without awaiting `ReceiveMeta::change_outpoint_range`. An application can therefore receive `await_final()`, immediately try to spend the promised credit, and get insufficient balance. Upstream's `await_receive` explicitly awaits these primary-module outputs for this reason.

Wait for those outputs before filling the credit fields/final state, preserving bounded `state()` behavior and the stream's shutdown handling. Test with mint issuance delayed after claim acceptance.

rust/fedimint-sdk/src/onchain/v1.rs
```diff
@@ -0,0 +1,799 @@
...
+    .map_err(|_| timeout())?
+    .map_err(|err| {
+        unreachable(format!(
+            "could not quote the withdrawal's on-chain fee: {err}"
+        ))
+    })?;
+    let output_value = sats_to_bitcoin(amount)
+        .checked_add(fees.amount())
+        .ok_or_else(|| internal("the withdrawal amount plus its on-chain fee overflowed"))?;
+    let quote = module.send_fee_quote(output_value).await.map_err(|err| {
```
zeenix: 🤖 ChatGPT (Codex) on behalf of zeenix:

[P2] Preserve insufficient-balance errors from the funding quote

When the balance covers the requested amount but not that amount plus fees (e.g. requesting the entire balance), `check_covers_amount` passes and `send_fee_quote` fails while selecting funding notes. This maps the failure to `Internal`; the later `available < plan.total` check is never reached. The same mapping exists in walletv2 and both send-time rechecks.

Return `InsufficientBalance` for this ordinary shortfall, as the Lightning facade already does, and cover an amount that fits before fees but cannot be funded after fees.

zeenix (commented):

Reviewed `36d087e`. Changes needed: five SDK implementation findings inline, including valid deposits becoming unobservable and cancelled receive subscriptions preventing clean shutdown.

Two existing upstream gaps also remain relevant:

- [fedimint#9101](https://github.com/fedimint/fedimint/issues/9101): walletv2 `receive()` still returns the same address until the scanner advances it. Two SDK receive calls mint different IDs for that same address and can both claim the same payment. The local IDs/address matching do not satisfy fresh-address reservation.
- [fedimint#9098](https://github.com/fedimint/fedimint/issues/9098): rechecking a funding quote and submitting are separate transactions. Concurrent spending or mint issuance can still change the actual debit after the check; the shared client read guard does not prevent this.

Keep those tracked upstream and add regression tests linked to the issues (temporarily ignored where necessary); they do not call for redesigning the agreed SDK API or duplicating upstream internals.

rust/fedimint-sdk/src/onchain.rs
zeenix: Fixed in 643384d. The shape check now requires exactly one wallet input and accepts any number of mint inputs beside it; the figures read the net mint movement (outputs minus inputs), charge the mint fee schedule on both sides, and derive dust from what the whole transaction funded less what it produced and paid. Unit tests cover a claim with mint inputs and the shape refusals, and the devimint round trip now deposits a second time into the wallet it just funded and withdrew from, so the claim the mint builds there can consolidate (8d29090); passes on both shapes.

rust/fedimint-sdk/src/onchain/v2.rs
zeenix: Fixed in 74c2225. The handle is scoped to the event-log scan and gone before the wait, and every client-owning read inside the receive stream (`find_link`, the final-state await, the issuance wait and the settlement) runs through `wait_holding_client` with the client's shutdown token, so a cancelled poll leaves nothing holding the client. Regression test `onchain_parked_deposit_subscriber_does_not_block_a_close` (8d29090) times out a second `next()`, keeps the subscriber and closes the federation within a bounded wait; passes on both shapes.

rust/fedimint-sdk/src/onchain/v1.rs
zeenix: Fixed in 643384d. `receive` no longer stores a placeholder in the entry's extra meta, and the `Deposit` backfill takes the address from upstream's own meta unconditionally, with `created_at` from the entry. A unit test replays exactly the crash-window payload (our old wire with an empty address in `extra_meta`) and checks the reconstructed record decodes with the real address.

rust/fedimint-sdk/src/onchain/v2.rs
zeenix: Fixed in 74c2225. After `Success` the stream waits until `has_active_states` is false for the claim's operation, which every state machine it registered, the mint output ones included, has to reach before the notes are spendable; only then are the credit fields filled and `Claimed` yielded. `Client::await_primary_module_outputs` is private at this pin, hence the stand-in, noted beside the code. `state()` keeps its bound and answers `Confirmed` until the wait is over. Mint issuance cannot be delayed from a test at this pin, so the round trip covers the ordinary case only.

rust/fedimint-sdk/src/onchain/v1.rs
zeenix: Fixed in 643384d and 03f3bc0. Both generations' `plan` and send-time re-check now classify a failed funding quote the way the lightning facade does: the typed insufficient-balance error, or the two mints' wordings matched against the whole error chain (the wallet's quote arrives wrapped in the client's submission error, which is why the downcast alone missed it). The round trip asks for the whole balance and expects `InsufficientBalance` (8d29090); passes on both shapes.

zeenix:
On the two upstream gaps: fedimint/fedimint#9101 has a regression test, `onchain_receive_hands_out_a_fresh_address_each_time`, ignored with the issue link; run with `--ignored` it fails on walletv2 today at the address comparison and passes on v1. fedimint/fedimint#9098 is noted at both send sites, beside the re-check, as the residual race it is; the window sits inside the submission itself, so no test from this side can drive it deterministically, and a test that cannot reproduce the defect would only document it in a place nobody runs. The comment carries the link instead.

Rebased tree at 03f3bc0: 501 unit / 16 integration (1 ignored) / 3 doc, devimint 16/16 on both v1 and walletv2.

rust/fedimint-sdk/src/onchain/v2.rs
zeenix: 🤖 ChatGPT (Codex) on behalf of zeenix:

The normal issuance delay is now handled, but `has_active_states == false` also includes terminal mint failures, so this still permanently reports `Claimed` when an output failed to become spendable (e.g. mintv2 `OutputSMState::Failure` after note verification fails).

There is already a public API at this exact pin: [`Client::await_primary_bitcoin_module_outputs`](https://github.com/fedimint/fedimint/blob/1ab15e4b89aaa35727ca16e401d0cbc7d066de58/fedimint-client/src/client.rs#L1391-L1402). It dispatches to the Bitcoin primary module and checks success versus failure for both mint generations. Please use it with `ReceiveMeta::change_outpoint_range.into_iter().collect()`, propagate an issuance failure without recording `Claimed`, and remove the polling workaround/comment claiming the needed API is unavailable. Keep the bounded snapshot and `wait_holding_client` handling.

zeenix (commented):

Rechecked `03f3bc0`. Resolved four findings: claim accounting with mint inputs, parked-subscriber shutdown, deposit-address backfill, and insufficient-balance mapping.

One correction remains in the issuance thread: use the existing public `await_primary_bitcoin_module_outputs` API. Polling for no active states cannot distinguish successful issuance from terminal failure.

The existing upstream gaps are now explicitly tracked in code/tests. CI is green.

rust/fedimint-sdk/src/onchain/v2.rs
zeenix: Fixed in 645ffc4. The stream now waits through `Client::await_primary_bitcoin_module_outputs` on the claim's `change_outpoint_range`; a `PrimaryModule` failure ends the deposit as `Failed` with the mint's own cause instead of `Claimed`, and any other failure of the wait is reported as an error rather than recorded. The polling loop, its interval constant and the comment about the private API are gone. The bounded read keeps its `has_active_states` snapshot so `state()` still answers `Confirmed` until the mint is done, and the wait still runs under `wait_holding_client`. devimint on walletv2 passes 16/16 (v1 code is untouched).

zeenix: 🤖 ChatGPT (Codex) on behalf of zeenix:

Verified in 645ffc4. The wait now checks the mint outputs' actual result: issuance failure becomes `Failed` before credit is recorded, and other wait errors propagate. The bounded state read and shutdown-safe subscription wait are preserved. This addresses the finding.

zeenix (commented):

Rechecked 645ffc4 against the pinned upstream implementation. The remaining mint-issuance finding is fixed; all five findings from my review are now resolved. No new blocking findings. Both PR workflows are green on this commit.

Ready from my side, with the previously discussed address-reservation and maximum-debit limitations still tracked separately in fedimint/fedimint#9101 and fedimint/fedimint#9098.

zeenix:
@fedimint-bot review

zeenix:
Fixed in 1ac4ce9. `receive` now reads the event log's tail before asking the module for the address and stores that earlier position, as upstream's doc on `WalletClientModule::receive` requires, so a claim the scanner makes in the gap after the address is handed out is not behind the cursor. devimint on walletv2 passes 16/16.

zeenix:
@fedimint-bot review

zeenix:
Addressed in 018f22d. The contract stays as documented, since the fresh-address-per-call gap is upstream's (fedimint/fedimint#9101, the ignored test tracks it), but the SDK now enforces the half it can: on walletv2, `receive` refuses with `Internal` when the module hands back an address an SDK record already owns, so an address is never handed out twice and two operations can never watch one address or adopt one claim. `Onchain::receive`'s error list names the outcome. New integration test `onchain_receive_never_hands_out_a_watched_address` runs on both generations: two calls never share an address, and the refusal is only accepted on walletv2, where it was confirmed to fire. devimint 17/17 on v1 and on walletv2.

zeenix:
@fedimint-bot review

zeenix:
Fixed in a07657a. The bounded read in `current_receive` now relinks past an aborted claim the way the subscription does: on `Aborted` it scans the event log from the aborted link's cursor, relinks the record to the retry's upstream operation when one is announced and checks that claim instead, so a caller that only polls `state()` follows the retry to `Claimed`. An aborted operation answers its final state at once, so the bound is only spent on the live claim at the end. An abort cannot be provoked from a test at this pin, so this is covered by the same round trip as before.

e65f77d fixes a gap in the earlier refusal of a repeated address, which the walletv2 round trip exposed: the module can hand a paid address back once more in the moment before its scanner advances, and the refusal now applies only while the owning operation is unfinished, since a finished owner's claim is already behind the new record's cursor. devimint 17/17 on v1 and on walletv2.

MrImmortal09:
@fedimint-bot review again

zeenix:
Fixed in c7d5634. walletv2's `receive` now runs under a per-federation deposit-allocation lock, from the event log position through the module's address, the owner check and the record commit, so two concurrent calls handed the same address record it exactly once and the other is refused. New integration test `onchain_concurrent_receives_never_share_an_address` races eight calls on separate tasks; with the lock removed it fails on walletv2 with one address recorded twice, with it in place devimint passes 18/18 on both v1 and walletv2.

MrImmortal09:
@fedimint-bot review again

zeenix:
Fixed in b175036, keeping the contract. walletv2's `receive` now refuses any address an SDK receive record already names, finished or not; a funded address is never handed out again. The reason a finished owner had been let through was the next deposit after a claim: walletv2 hands its highest derived address back until the scanner sees it paid and only then grinds the next one, so the call right after a claim got the old address. Instead of letting it through, `receive` now waits for the scanner's next address within `CONTACT_TIMEOUT`, the way the first `receive` on a fresh wallet waits for the scanner's first address, and fails with `Timeout` past that bound. The unfinished-owner case is still refused at once. `Onchain::receive`'s error list describes both. devimint 18/18 on v1 and on walletv2, including the round trip's second deposit, which exercises the wait.

MrImmortal09 (approved):
LGTM

PR #394, opened 2026-09-16
feat(sdk): runnable examples, and an exact AnyOperation::support (T16)

zeenix:
## Description

- ref fedimint-sdk-implementation-plan T16; closes the `AnyOperation::support()` gap #377 left open, now that #392 landed the last driver arm
- `support()` is exact: the driver lookup is keyed by `OperationKind` with an exhaustive match and no wildcard, so a kind this build recognises has a driver by construction, and a unit test walks every kind tag to check it. Every comment describing the old gap is gone
- four runnable examples under `rust/fedimint-sdk/examples/`, each an ordinary program over a data directory and an invite code that works against any federation: `walkthrough` (the crate doc's walkthrough made real), `ecash` (`send`, `receive`, `cancel`), `lightning` (`receive`, `send`) and `onchain` (`receive`, `send`). The data directory persists between runs, which is how a send finds the balance an earlier receive brought in and how a stored operation is picked up again by id. Anything a counterparty must do (pay an invoice, send coins, redeem notes) the example prints as a `key: value` line and waits for. They use the public API only, and `cargo test` compiles them and clippy lints them on every CI run
- the examples know nothing about devimint. `scripts/run-sdk-examples.sh [v1|v2] [example ...]` stands a devimint federation up and plays the counterparty: it starts an example, picks the invoice, address or notes out of its output, pays or sends from devimint's faucet and bitcoind, and waits for the example to finish. The shape selection moved into `scripts/devimint-shape.sh`, sourced by both runner scripts. The default shape is `v2`: at the pinned revision the v1 lightning client cannot decode a successful send's preimage (fedimint/fedimint#8969), so on `v1` the `walkthrough` and `lightning` examples end with an error after the payment went through
- the crate docs and README no longer call the crate a skeleton; the doc walkthrough is a short snippet pointing at the example
- running the examples found a defect: a mintv2 ecash receive reported `Done` before its notes were spendable, so a balance read at that moment was short by the credit. The SDK now waits on the mint's outputs before `Done` (and before a reclaim's `Canceled`), the same shape the on-chain claim got in #392. Separate commit, easy to split out if wanted

## Testing

- 502 unit tests (1 new: every kind tag has a driver and reads `Observable`), 18 integration, 3 doc; fmt, clippy `--all-targets -D warnings` (examples included), rustdoc, wasm32 check; shellcheck clean on the scripts
- under devimint through the runner: all four examples pass on walletv2, with the ecash notes sent by one wallet and redeemed by a second, the sender's cancellation then reading `Redeemed`; `ecash` and `onchain` pass on v1, `lightning` and `walkthrough` fail on v1 with exactly the documented preimage error
- the integration suite is 18/18 on both shapes


rust/fedimint-sdk/examples/ecash.rs
```diff
@@ -0,0 +1,112 @@
...
+            let received = ecash.receive(&notes).await?;
+            let mut updates = received.updates();
+            while let Some(state) = updates.next().await? {
+                println!("state: {state:?}");
+            }
+            let details = received.details().await?;
+            println!(
+                "redeemed {} of notes minus {} fee ({} credited)",
+                details.notes_value, details.fee, details.net_credit,
+            );
```
zeenix: [P2] Check the terminal state before reporting a credit

`updates.next()` also ends normally after `EcashReceiveState::Failed`. For example, if the sender reclaims the notes first, this still prints “redeemed … credited” using the amounts recorded before redemption succeeded. Only print that receipt for `Done`. The same issue occurs in `lightning::receive` after `Expired`/`Canceled`/`Failed`; the on-chain receive instead panics on the missing fee/credit after `Failed`. Please handle the terminal outcome explicitly in all three examples.

— Codex, reviewing on behalf of zeenix.

scripts/run-sdk-examples-inner.sh
```diff
@@ -0,0 +1,246 @@
...
+
+# Receives a payment, then sends a fresh faucet invoice back out.
+run_lightning() {
+  local receive_log="$root/lightning-receive.log"
+  start_example lightning "$receive_log" "$wallet_a" "$invite" receive 200000 "an example"
+  local invoice
+  invoice="$(grab "$receive_log" "invoice:" "$pid")" || die lightning
+  pay_invoice "$invoice"
+  finish lightning "$pid" "$receive_log"
+  run_example lightning "$root/lightning-send.log" "$wallet_a" "$invite" send "$(new_invoice 50000)"
```
zeenix: [P2] Assert the payment outcome in the runner

The examples intentionally exit successfully for observed `Refunded` and `Failed` states, so a successful `run_example` does not establish that this payment worked. Running this actual script with simulated example outputs confirms that paid, refunded, and failed sends all produce exit status 0. Check the expected terminal outcome here and for the walkthrough/on-chain sends, as the ecash cancellation already does, so payment regressions fail the runner.

— Codex, reviewing on behalf of zeenix.

scripts/run-sdk-examples-inner.sh
```diff
@@ -0,0 +1,246 @@
...
+    fi
+    sleep 1
+    waited=$((waited + 1))
+  done
+  return 1
+}
+
+# Pays a bolt11 invoice from a node outside the federation: the counterparty for a receive.
+pay_invoice() {
+  curl -sS -X POST --data "$1" "http://localhost:$FM_PORT_FAUCET/pay" >/dev/null
```
zeenix: [P2] Propagate faucet HTTP failures

`curl -sS` returns success for HTTP 4xx/5xx responses. If the faucet rejects the payment, this discards its error body and proceeds to `finish`, which waits without a timeout for an invoice that was never funded. Use `--fail-with-body` (or check the HTTP status), and route the failure through `die` with the example name. Apply the status check to `new_invoice` too.

— Codex, reviewing on behalf of zeenix.

rust/fedimint-sdk/examples/onchain.rs
```diff
@@ -0,0 +1,105 @@
...
+                quote.fee(),
+                quote.fee_breakdown(),
+                quote.total(),
+                quote.expires_at(),
+            );
+            let send = onchain.send(quote).await?;
+            let outcome = send.await_final().await?;
+            println!("state: {outcome:?}");
+            if let OnchainSendState::Succeeded { txid } = outcome {
+                println!("confirmed on chain, txid {txid}");
```
zeenix: [P2] Describe this transaction as broadcast

`OnchainSendState::Succeeded` explicitly means the federation broadcast the transaction; Bitcoin confirmations happen afterwards and are not tracked by this operation. This line therefore reports confirmation while the withdrawal can still be unconfirmed. Print “broadcast” here, or independently verify confirmation before claiming it.

— Codex, reviewing on behalf of zeenix.

scripts/run-sdk-examples.sh
```diff
@@ -0,0 +1,46 @@
...
+# error after the payment has actually gone through, while ecash and onchain are unaffected.
+#
+# Must run inside `nix develop --accept-flake-config .#wasm-tests`: that is the
+# only dev shell with devimint, fedimintd, gatewayd, bitcoind, lnd, esplora and
+# the recurringd binaries on PATH.
+
+set -euo pipefail
+
+shape="${1:-v2}"
+[ $# -gt 0 ] && shift
```
zeenix: [P3] Preserve the first example when the shape is omitted

The documented interface makes the shape optional, but `scripts/run-sdk-examples.sh lightning` treats `lightning` as the shape and exits 2. Only consume the first argument when it names a supported shape; otherwise retain the v2 default and pass all arguments through as example names.

— Codex, reviewing on behalf of zeenix.

rust/fedimint-sdk/src/ecash.rs
```diff
@@ -1507,22 +1507,44 @@ async fn mintv2_reclaim(
...
-    Ok(match outcome {
-        fedimint_mintv2_client::FinalReceiveOperationState::Success => EcashSendState::Canceled,
-        fedimint_mintv2_client::FinalReceiveOperationState::Rejected => EcashSendState::Redeemed,
-    })
+    match outcome {
+        Mintv2ReceiveOutcome::Done => Ok(EcashSendState::Canceled),
+        Mintv2ReceiveOutcome::Rejected => Ok(EcashSendState::Redeemed),
+        // The reclaim was accepted, but never became spendable, so it settled nothing: the
+        // notes it targeted are still exactly as redeemed or not as they were before this
+        // ran, and the caller needs to know the reclaim itself did not actually happen.
```
zeenix: [P3] Correct the new settlement explanations

An accepted reclaim has already spent its input notes, even if issuing the replacement notes fails; the original notes are not unchanged or still reclaimable. Please describe that distinction here. The new comment in `EcashReceiveDriver::current` also misstates the pinned implementation: `await_primary_bitcoin_module_outputs` reads the mint's local state notifications, and `has_active_states(id)` includes its output state machines. It does not inherently require another federation round trip after those machines have finished.

— Codex, reviewing on behalf of zeenix.

zeenix (commented):
Reviewed by Codex on behalf of zeenix at `3890748`. Findings are inline.

Commit hygiene: `a5486ac` combines the receive-state refactor with unrelated crate-allowance and shell-comment edits; please split those into their respective logical changes. No persisted-format change identified.

rust/fedimint-sdk/examples/ecash.rs
zeenix: Fixed in ccec112. All three receive subcommands keep the last state the update stream yielded and print the receipt only on the success variant (`Done`, `Claimed`); any other terminal state prints one line saying the receive did not complete and which state it ended in, then the balance. The on-chain example no longer reads the fee and credit after a `Failed`, so it cannot panic there.

scripts/run-sdk-examples-inner.sh
zeenix: Fixed in daad2f2. Every sequence now checks its log for the expected terminal line the way the ecash cancellation did: `state: paid, ` for the lightning send and the walkthrough's payment (plus a `notes:` line for its ecash send), `state: Claimed` for the lightning receive, the ecash funding and the on-chain receive, `state: Done` for the ecash receive, `state: Succeeded` for the on-chain send; each failure goes through `die` with the example's name. Verified live on both shapes.

scripts/run-sdk-examples-inner.sh
zeenix: Fixed in daad2f2. Both faucet calls use `curl -sS --fail-with-body`, and a failure is routed through `die` with the name of the example the sequence was running, so a rejected payment or a refused invoice stops the run instead of leaving `finish` waiting on an invoice nobody funded.

rust/fedimint-sdk/examples/onchain.rs
zeenix: Fixed in 722900f. The line reads `broadcast, txid ..` now, matching what `Succeeded` promises.

scripts/run-sdk-examples.sh
zeenix: Fixed in daad2f2. The first argument is taken as the shape only when it is `v1` or `v2`; otherwise the default stays `v2` and every argument is passed through as an example name, so `scripts/run-sdk-examples.sh lightning` runs the lightning example.

rust/fedimint-sdk/src/ecash.rs
zeenix: Fixed in 79a646a. The `NotIssued` comment in `mintv2_reclaim` now says the accepted reclaim already spent its input notes and only the issuance of their replacements failed, so the value is neither back in the balance nor still reclaimable, which is what the error tells the caller. The comment in `EcashReceiveDriver::current` now says the outputs wait is served from the client's own state-machine notifications, that `has_active_states` already covers the output state machines and is the bound's guard, and that `now_or_never` only guards against the wait not being immediately ready.

zeenix:
On the commit hygiene: a5486ac is split into its three logical changes, 567d962 (the shared `mintv2_receive_state` mapping), a78bbbf (the crate-level allowance comment) and 66da239 (the runner's comment on the client command line); the branch was rebuilt on top of them and its tree is byte-identical to before the split. The six inline findings are addressed in ccec112, 722900f, 79a646a and daad2f2, each answered in its thread. Under devimint the runner passes all four examples on walletv2 and ecash and on-chain on v1, with the new outcome checks in place.

rust/fedimint-sdk/examples/ecash.rs
```diff
@@ -0,0 +1,122 @@
...
+            let received = ecash.receive(&notes).await?;
+            let mut updates = received.updates();
+            let mut last_state = None;
+            while let Some(state) = updates.next().await? {
+                println!("state: {state:?}");
+                last_state = Some(state);
+            }
+            // `Done` is the only terminal state whose notes became spendable: `Failed`
+            // means they were already spent or reclaimed by whoever sent them, so
+            // there is no receipt to show.
```
zeenix: [P3] Keep the failure explanation broad enough for issuance failures

`Failed` does not necessarily mean the sender spent or reclaimed the notes: this PR's `mintv2_receive_state` also produces it for `NotIssued` after an accepted redemption. Please say that the receive did not produce spendable notes, or use “most often” as the enum documentation does, so the example does not teach callers to infer a double spend from every failure.

— Codex, reviewing on behalf of zeenix.

scripts/run-sdk-examples.sh
```diff
@@ -0,0 +1,51 @@
...
+# only dev shell with devimint, fedimintd, gatewayd, bitcoind, lnd, esplora and
+# the recurringd binaries on PATH.
+
+set -euo pipefail
+
+shape=v2
+if [ $# -gt 0 ] && { [ "$1" = v1 ] || [ "$1" = v2 ]; }; then
+  shape="$1"
+  shift
+fi
```
zeenix, line 31: [P3] Split the new runner commit into its logical fixes

Could you split `daad2f2` into faucet HTTP error handling, payment-outcome assertions, and optional-shape parsing? These fix three independent defects and can be reviewed/reverted separately. The previous `a5486ac` split is correct; the same atomic-commit rule applies to this new commit.

— Codex, reviewing on behalf of zeenix.

zeenix (commented):
Rechecked `daad2f2` by Codex on behalf of zeenix. All six prior findings are addressed; those threads are resolved. The earlier commit split preserves the exact tree. CI is green, and 20 targeted shell cases passed using simulated example/faucet responses. No new functional defects found; two minor cleanup items remain inline.

rust/fedimint-sdk/examples/ecash.rs
zeenix: Fixed in 0f88f0d. The comment now says a `Failed` receive produced no spendable notes, most often because the notes were already spent or reclaimed by the sender, matching the state's own documentation instead of naming a double spend as the only cause.

scripts/run-sdk-examples.sh
zeenix: Done. daad2f2 is now three commits: e6e2891 (the shape argument is consumed only when it is `v1` or `v2`), 4c29f57 (`--fail-with-body` on both faucet calls, routed through `die`, with the lightning send's invoice fetched into a variable first so a failure there is not swallowed inside a command substitution) and d0fe7d4 (the per-sequence outcome assertions). The tree after the third commit is byte-identical to daad2f2, so the runner results reported earlier still hold.

zeenix (commented):
Rechecked `0f88f0d` by Codex on behalf of zeenix. All findings are addressed and the remaining threads are resolved. The three runner fixes are now separate commits with an identical resulting tree; the only subsequent change corrects the ecash comment. The previously verified runner behavior is unchanged, CI is green, and I have no remaining findings. Ready to merge from this review.

MrImmortal09:
@fedimint-bot review.

zeenix:
Fixed in 3b0ec13. The examples no longer print the seed. The first run over a fresh data directory prints one line on stderr saying a new wallet was created and that `Sdk::export_mnemonic` returns the phrase to back up, without the phrase itself, and the comment there says why it stays out of the output. No runner sequence or documented line depended on the old `mnemonic:` line.

MrImmortal09:
@fedimint-bot review again.

zeenix:
I did some manual testing of the examples too and it seems to works fine:

Screenshot From 2026-09-17 16-53-03

Found a few issue, so had claude fix them. One of them was in the SDK itself so that's [a separate PR](https://github.com/fedimint/fedimint-sdk/pull/396).

zeenix:
@MrImmortal09 Given the clankers are happy and I tested manually as well, I think it's safe to merge so I'm doing that now.

PR #395, opened 2026-09-16
ci: wire devimint integration tests to mapped workflow (T13)

Ayush4958:
### Description
Integrates the Rust SDK devimint test suite into the main `pull-request.yml` workflow using the `detect-changes` pattern to prevent PR deadlocks. 

Key architectural details:
- **Global Concurrency Lock**: Serializes all devimint runs across the self-hosted runner to completely eliminate port collisions.
- **Ephemeral Workspaces**: Configures `devimint` to write to the runner's `$GITHUB_WORKSPACE` instead of `/tmp`, preventing gigabytes of artifact bloat from past runs.
- **Non-blocking Rollout**: Marked `continue-on-error: true` so the suite builds a reliability track record without blocking team velocity on day one.


Ayush4958:
@zeenix @MrImmortal09 
PR is ready for review now

Thanks ...

MrImmortal09:
@fedimint-bot review

Ayush4958:
@MrImmortal09 
PTAL again 

thanks  ....

MrImmortal09:
@fedimint-bot review

.github/workflows/pull-request.yml
```diff
@@ -146,6 +150,10 @@ jobs:
               flake.nix|flake.lock)
                 web=true
                 react_native=true
+                rust_sdk=true
+                ;;
+              rust/fedimint-sdk/**|scripts/run-sdk-integration-tests.sh)
```
zeenix: Posted by Codex on behalf of @zeenix.

**[P2] Match the SDK test script before the generic scripts rule**

The earlier `scripts/*` arm matches this path and terminates with `;;`, so this arm is never reached
for `scripts/run-sdk-integration-tests.sh`. A PR changing only that script produces
`rust-sdk=false` and skips both SDK integration jobs. I reproduced this using the workflow's case
statement. Move the specific script match before the generic rule, or set both affected outputs
in the earlier arm.

.github/workflows/pull-request.yml
```diff
@@ -221,3 +229,56 @@ jobs:
...
+    timeout-minutes: 45
+    # Non-blocking on day one: let the suite build a track record before
+    # promoting it to a required check.
+    continue-on-error: true
+    concurrency:
+      # Global lock: serializes all devimint runs on the
+      # self-hosted runner regardless of which PR or matrix leg triggered them,
+      # so two concurrent jobs cannot collide on ports or state.
+      group: devimint-integration
+      cancel-in-progress: false
```
zeenix, line 247: Posted by Codex on behalf of @zeenix.

**[P2] Preserve queued matrix jobs in the global concurrency group**

Without `queue: max`, GitHub permits only one pending job in this group. For example, while PR A's
v1 job runs and its v2 job waits, PR B's matrix jobs replace pending jobs instead of waiting behind
them. `cancel-in-progress: false` protects the running job only, so unrelated PRs lose integration
coverage. Add `queue: max` to retain the global serialization while allowing pending jobs to queue.

scripts/run-sdk-integration-tests.sh
```diff
@@ -111,4 +111,32 @@ grep -m1 -o 'rev = "[0-9a-f]\{40\}"' Cargo.toml
...
+MAX_RETRIES=3
+BASE_TEST_DIR="${FM_TEST_DIR:-/tmp/devimint-tests}"
+
+for ((i=1; i<=MAX_RETRIES; i++)); do
+  echo "Running integration tests (Attempt $i of $MAX_RETRIES)..."
+
+  # Devimint uses the exact path in FM_TEST_DIR. We must give each attempt
+  # a completely clean directory, otherwise bitcoind wallet state from the
+  # failed attempt will corrupt the retry.
+  export FM_TEST_DIR="$BASE_TEST_DIR/attempt-$i"
```
zeenix, line 129: Posted by Codex on behalf of @zeenix.

**[P2] Allocate a unique federation directory for each script invocation**

Every invocation starts again at `$BASE_TEST_DIR/attempt-1`, and `mkdir -p` preserves any existing
state. With no override, separate local runs all share `/tmp/devimint-tests/attempt-1`; concurrent
runs share it too. This replaces devimint's previous unique default directory and can reuse stale
federation configuration or collide on daemon databases. A two-invocation shell probe confirmed
that the second run sees state left by the first. Allocate a unique directory per invocation
(e.g. with `mktemp -d`), then put the retry directories underneath it.

zeenix (changes requested):
Posted by Codex on behalf of @zeenix.

Three P2 issues need fixing: the script path filter is shadowed, the global concurrency group
discards pending jobs, and separate script invocations reuse federation state.

Validated the filter and directory reuse with focused shell probes. Both current CI shapes passed
18 tests with 4 ignored; those runs do not exercise these cases.

Ayush4958:
@zeenix 
Fixes all addressed reviews.

PTAL again 
thanks ...

zeenix (approved):
Posted by Codex on behalf of @zeenix.

Rechecked 6568ef8: all three findings are addressed; no new defects found.

Verified the script-specific filter, `queue: max`, unique default directories across sequential
and concurrent invocations, and isolated retry directories under the CI log path. Both current
CI shapes passed on their first attempt: 18 tests passed, 4 ignored each.

PR #396, opened 2026-09-17
fix(sdk): read a fee quote refusal through the client wrapper; too-small receives are InvalidInput

zeenix:
## Description

- a lightning receive too small to cover its claim fee came back as `Internal` with the message `could not quote the claim fee: The primary module failed`. Seen on walletv2 with a 200 msat receive on a fresh wallet: lnv2's claim fee has a 1 sat base, so the contract can never cover it
- the client wraps the mint's refusal in an error whose own message is only "The primary module failed". The SDK's fee-quote classifier read that outer message alone, so it never saw the mint's `Insufficient funds` (mintv2) or `Insufficient balance` (mint v1) underneath and fell through to `Internal`. The send paths mostly hid this behind their early balance check; the receive paths have none
- the classifier now walks the whole error chain, and the `Internal` fallback carries the chain's full text. The mint v1 typed error is matched by its message too: the client re-boxes the mint's error through anyhow, which keeps `Display` and `source` but drops the concrete type, so the downcast cannot succeed on the shape the client produces today. A unit test builds that exact shape, with `anyhow` as a dev-dependency only; the library still never names it
- a receive quote refused for insufficient funds means the claim fee exceeds the contract amount, which is an amount problem: it is now `InvalidInput`, "the amount is too small to cover the fees of claiming it", the same error a funded wallet gets from the net-credit check. One constructor serves every place that says it. Sends keep `InsufficientBalance`
- the v1 lightning module's claim fee is zero at the pinned revision, so the new integration test expects `Ok` there and `InvalidInput` on v2

## Testing

- 530 unit (1 new: both mints' refusals are recognised through the client's wrapper, and the fallback carries the chain), 19 integration (1 new), 3 doc; fmt, clippy `--all-targets`, rustdoc, wasm32 check
- devimint integration suite 19/19 on v1 and v2. One v2 attempt failed the on-chain round-trip test on its claim timeout and passed on the retry; unrelated to this change


zeenix:
@fedimint-bot review

rust/fedimint-sdk/src/lightning.rs
```diff
@@ -906,63 +907,122 @@ pub(super) fn plan_of(
...
-/// mint's answer carries no amounts of its own. `context` names the quote for the fallback
-/// message, when `short` is absent and `text` does not match either mint's wording for "the
-/// notes on hand are short".
+/// `shortfall` says what an insufficient-balance refusal means at the caller's call site.
+/// `required` is the amount the failed quote was for, used to report the shortfall when
+/// `shortfall` is [`Shortfall::Balance`] and the mint's answer carries no amounts of its own.
+/// `context` names the quote for the fallback message, when neither mint's refusal is found
+/// anywhere in `err`'s chain.
 pub(super) async fn fee_quote_failure(
     client: &Client,
```
zeenix, line 1004: **[P1] Preserve `Send + Sync` on the error parameter**

This async function captures `&dyn Error`, which is not `Send` because `dyn Error` lacks
`Sync`. This propagates to `Lightning::quote`, `send`, and `receive`, so callers can no longer
await any of them inside `tokio::spawn`, even when the operation succeeds.

Use `&(dyn std::error::Error + Send + Sync + 'static)` here. I confirmed compile-time `Send`
assertions for all three public futures fail at this head and pass with that one-line change.
Please retain those assertions as native regression tests.

zeenix (commented):

One blocking finding: the new async helper makes all three Lightning futures non-`Send`.
The fee classification otherwise matches the pinned upstream implementations.

rust/fedimint-sdk/src/lightning.rs
zeenix: Fixed in 8447d70. The parameter is `&(dyn std::error::Error + Send + Sync + 'static)` now, and a compile-time test names the `quote`, `send` and `receive` futures behind a `Send` bound; with the old signature it fails to compile with one error per future, with the new one it passes. Gates and the wasm check are green.

zeenix: Rechecked by Codex on behalf of @zeenix at 8447d70.

Confirmed fixed. The helper preserves `Send + Sync`, and the new compile-time regression test
covers all three public futures. `cargo check --locked --offline --features uniffi --tests`
passes locally; the Rust SDK CI also passes. No new findings in the update. Resolving this blocker.

zeenix:
Same here, @MrImmortal09 . Two clankers are happy and this is a fix of a bug.

PR #400, opened 2026-09-19
fix(workflows): pass changeset flags to the commands that accept them

### LINKED ISSUE #397: CI: Changesets canary publish passes removed --snapshot flag
The `Changesets` workflow is failing on `main` after the `@changesets/cli` v3 upgrade because the canary publish command still passes `--snapshot` to `changeset publish`.

Failed run: https://github.com/fedimint/fedimint-sdk/actions/runs/35250873204
Head SHA: `a02c387f2a256621157bbea63ca44998ab1da3a7`
Failed job/step: `Publish Changesets` / `Publish prerelease` first, then `Upload Artifacts to Release (Canary)` also failed while replacing the release asset.

Focused log excerpt:

```text
$ changeset publish --no-git-tag --snapshot canary --tag canary
changeset v3.0.1
CACError: Unknown option `--snapshot`
[ELIFECYCLE] Command failed with exit code 1.
```

The later release-upload failure was:

```text
Uploading ios-artifacts.zip...
error: other side closed
```

I checked the landing commit/PR for the failed run: #396 only touched Rust SDK lightning code/tests, while all verify/build/test jobs passed. This is a repository-level release workflow issue, not fallout from #396.

Local validation:

```text
pnpm --dir js changeset publish --help
changeset v3.0.1
Options:
  --otp <code>
  --tag <name>
  --from-pack-dir <dir>
  --git-tag
  -h, --help
```

`changeset version --help` still supports `--snapshot [name]`, so the snapshot flag should stay on `changeset version` and be removed only from `changeset publish`.

Exact patch I prepared locally:

```diff
diff --git a/.github/workflows/changesets.yml b/.github/workflows/changesets.yml
@@
-            pnpm --dir js changeset publish --no-git-tag --snapshot canary --tag canary
+            pnpm --dir js changeset publish --no-git-tag --tag canary
diff --git a/.github/workflows/snapshot.yml b/.github/workflows/snapshot.yml
@@
-          pnpm --dir js changeset publish --no-git-tag --snapshot \$snapshot --tag snapshot
+          pnpm --dir js changeset publish --no-git-tag --tag snapshot
```

Checks run locally:

```text
pnpm --dir js lint:fix
pnpm --dir js changeset publish --help
git diff --check
rg -n "changeset publish .*--snapshot|--snapshot .*changeset publish" .github/workflows -S
```

All passed. The runner reports the existing Node engine warning (`current: node v24.19.0`, repo wants `>=22.12.0 <23`).

I attempted to push a draft-PR branch named `fedimint-bot/ci-debug-changeset-snapshot-35250873204`, but GitHub rejected it because this bot token cannot update workflow files without `workflow` scope:

```text
refusing to allow a Personal Access Token to create or update workflow `.github/workflows/changesets.yml` without `workflow` scope
```

Concrete next action: apply the two-line workflow patch above with a token/user that has workflow-file permissions, then rerun the failed Changesets workflow.

Yunusabdul38:
## The change

Three things, across `changesets.yml` and `snapshot.yml`:

**1. Flags moved to the commands that accept them**

```diff
-changeset version --no-git-tag --snapshot canary
+changeset version --snapshot canary

-changeset publish --no-git-tag --snapshot canary --tag canary
+changeset publish --no-git-tag --tag canary
```

**2. Canary failures are no longer swallowed** — `continue-on-error: true` removed, `set -euo pipefail` added (matching `snapshot.yml`).

**3. A guard so "nothing to release" isn't reported as a failure** — see below.

## A stricter argument parser stopped ignoring the flags

The `@changesets/cli` v3 bump (#366) replaced the `mri` parser with `cac`, which validates options per command instead of ignoring unknown ones:

```
$ changeset version --no-git-tag --snapshot canary
CACError: Unknown option `--gitTag`

$ changeset publish --no-git-tag --snapshot canary --tag canary
CACError: Unknown option `--snapshot`
```

`--snapshot` is declared only on `version`, `--git-tag` only on `publish` — in v2 as well as v3, so nothing that ever worked is being removed. `version` applies the snapshot name; `publish` only needs `--tag`.

## #397 reports half the bug

**Both commands fail, not just `publish`**, so the patch in the issue would still have died on the first line. The `version` failure is missing from the logs because the canary step had no `set -e`: bash continued to `publish`, and only that error surfaced.

The run linked in the issue also went red on `Upload Artifacts to Release (Canary)`, not on the changesets step (which was `continue-on-error: true`). That step is gone since #388, so this is the last piece.

## Two weeks of canaries never shipped

Canary publishing has been failing silently since 2026-09-05. The latest canary on npm is from 2026-08-28, 19 commits ago, so anyone testing against `@fedimint/core@canary` is pinned to late-August code.

## Making it loud, without crying wolf

Per review, the step now fails the workflow rather than failing quietly. One wrinkle had to be handled first: `changeset version --snapshot` exits 1 when no changeset is pending

```
No unreleased changesets found.
🦋 Exited with code 1
```

which is the ordinary state of any commit that needs no release (this PR included). `set -e` plus removing `continue-on-error` alone would therefore redden `main` on every changeset-less merge, and an alarm that fires on nothing gets muted as fast as one that never fires.

So each workflow now handles that case explicitly, and differently, because the two are triggered differently:

- **`changesets.yml`** fires on every push, and nobody asked it for a canary: with no changeset pending it logs the reason and exits 0.
- **`snapshot.yml`** is run by a person asking for a snapshot: with no changeset pending it still fails, but says why in plain words and does it before the install rather than after a full build.

```
No changeset pending on this branch; there is nothing to publish as a snapshot.
```

Both guards carry a comment explaining the difference, so neither reads as an oversight. 

## Verification

Neither workflow has a `pull_request` trigger, so CI here can't exercise them — prettier is the only check that touches these files. I ran `@changesets/cli` 3.0.1 against a scratch workspace using this repo's `.changeset/config.json`, with a `private` package and npm pointed at a dead registry:

```
canary flow (changesets.yml)
  changeset pending → 0.0.0-canary-<sha> → publish   exit 0  green
  none pending      → "No changeset pending…"        exit 0  green

snapshot flow (snapshot.yml)
  changeset pending → 0.0.0-<branch>-<sha> → publish exit 0  green
  none pending      → "No changeset pending…"        exit 1  red

a real failure mid-script (build breaks)
  before → publish still ran, step exit 0
  after  → stops at the failure, step exit 1
```

Before the fix, the first line of each flow was `CACError`.

Also checked: `pnpm --dir js lint` passes, both files parse as YAML, `bash -n` is clean on both inline scripts, and no step in either file carries `continue-on-error` any more.

After merge, `npm view @fedimint/core dist-tags` should show a `canary` entry ending in the new commit's SHA.

Fixes #397


MrImmortal09:
@fedimint-bot review

MrImmortal09:
> ## Worth a separate look: the step that fails quietly
> `continue-on-error: true` on the prerelease step is why this went unnoticed for 19 commits. Removing it may be wrong — a transient npm failure shouldn't redden `main` — but a release step failing with no signal seems worth revisiting. Happy to open a separate issue.
> 


Lets have this done in the same PR. Correct me if I am wrong, I feel we should throw error and exit if its safe to

Yunusabdul38:
> Lets have this done in the same PR. Correct me if I am wrong, I feel we should throw error and exit if its safe to

You're right that it should be loud, and I've done it in this PR. One wrinkle turned up while testing, though.

`changeset version --snapshot` exits 1 when no changeset is pending:

```
$ changeset version --snapshot canary
No unreleased changesets found.
🦋 Exited with code 1
```

That's the ordinary state of any commit that doesn't need a release — this PR included. So `set -e` plus dropping `continue-on-error` on their own would turn `main` red on every changeset-less merge, and an alarm that fires on nothing gets muted about as fast as one that never fires.

So the canary step now tells the two cases apart:

- `set -euo pipefail`, matching `snapshot.yml`
- `continue-on-error: true` removed
- skips early with a log line (exit 0) when no changeset is pending

A bad flag, a build break or an npm failure now fails the workflow. "Nothing to publish" doesn't.

Tested against `@changesets/cli` 3.0.1 in a scratch workspace:

```
commit WITH a pending changeset:   before → exit 0      after → exit 0 (canary published)
commit WITHOUT one:                before → exit 1 RED  after → exit 0 ("no changeset pending; skipping")
```

`snapshot.yml` gets the same condition handled explicitly, but deliberately keeps failing rather than skipping: that workflow is triggered by a person asking for a snapshot, so "there is nothing to snapshot" is an answer they want, not noise on `main`. What changes is that it now says so in plain words and fails in seconds, instead of running a full install and build before surfacing a changesets error:

```
No changeset pending on this branch; there is nothing to publish as a snapshot.
```

Both guards are commented with why they differ, so neither reads as an oversight.

MrImmortal09:
@fedimint-bot review

MrImmortal09 (approved):
LGTM

PR #401, opened 2026-09-20
feat(ios): init Swift SDK, XCFramework pipeline and demo app (T15)

sadeeqrabiu:
Introduces the Swift/iOS SDK for fedimint-sdk, mirroring the two-stage, zero-drift architecture established for Android Kotlin in #388:

  1. The native libraries: cross-compiled for Apple targets via plain cargo (aarch64-apple-ios, aarch64-apple-ios-sim, x86_64-apple-ios, and host aarch64-apple-darwin) and packaged into FedimintSdkFFI.xcframework.
  2. The Swift bindings: generated by uniffi-bindgen directly from the metadata embedded in the native static archive (libfedimint_sdk.a), guaranteeing the Swift API cannot drift from the device binary.

Key highlights & design decisions:
- Swift Package Manager (ios/Package.swift): distributes FedimintSdk as a modern Swift package backed by the binary XCFramework target.
- C++ runtime linkage: wires .linkedLibrary("c++") to satisfy rocksdb and aws-lc runtime dependencies across Apple platforms.
- Error naming workaround: works around UniFFI 0.32 constructor rename limitation by providing public typealias SdkError in Version.swift.
- Framework linking: adds SystemConfiguration, Security, and Network to uniffi.toml link_frameworks for iOS socket and network monitor symbols.
- Demo app (ios/Demo): single-screen SwiftUI dashboard driving wallet lifecycle, balance streaming, Ecash, Lightning, Onchain, and Activity. Configured via committed project.yml (xcodegen).
- CI workflow (.github/workflows/swift-sdk.yaml): builds XCFramework, runs offline package tests on macOS slice, and compiles demo app for simulator.

Co-Authored: Claude Opus 5


MrImmortal09:
@fedimint-bot review

MrImmortal09:
@fedimint-bot review again

MrImmortal09:
@fedimint-bot review again , this time start a fresh run

ios/Demo/DemoModel.swift
```diff
@@ -0,0 +1,593 @@
...
+                let updates = joined.balanceUpdates()
+                while !Task.isCancelled {
+                    let next = try await updates.next()
+                    // Scoped to one iteration, so nothing strong is held across
+                    // the await above. `break` rather than skipping: with the
+                    // model gone there is nobody left to render into, and
+                    // `next()` is a uniffi call that cancellation cannot stop.
+                    guard let self else { break }
+                    self.balance = "Balance: " + formatMsats(next)
+                }
```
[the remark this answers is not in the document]
sadeeqrabiu: Exactly right. Addressed in [5f45431](https://github.com/fedimint/fedimint-sdk/commit/5f45431), both post-await balance writes and the catch blocks are now guarded with !Task.isCancelled so an in-flight UniFFI call from a canceled task cannot overwrite or stamp errors over the replacement federation.

MrImmortal09:
@fedimint-bot review again , this time start a fresh run

.github/workflows/swift-sdk.yaml
```diff
@@ -0,0 +1,145 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  swift:
+    name: 'Build the XCFramework, generate the bindings, test + compile the demo'
+    # Apple Silicon. The macOS SDK it ships is what every slice here compiles
+    # against, so this cannot move to a Linux runner, Nix or not.
+    runs-on: macos-15
```
MrImmortal09, line 59: Any reason for macos-15 and not any of the latest one ?

MrImmortal09:
@fedimint-bot review again, and check if any of the past review points is missing to be completed.

MrImmortal09 (changes requested):
Hey @sadeeqrabiu,                                                                                                                                                            
                                                                                                                                                                                                            
  One quick note on caching: take a look at how we handled the iOS build in fedimint/fedimint-sdk-ffi flake.nix https://github.com/fedimint/fedimint-sdk-ffi/blob/main/flake.nix#L110-L180.                 
                                                                                                                                                                                                            
  By using __noChroot = true (relaxed sandbox) and a small xcode-wrapper symlink derivation, we were able to build the iOS slices as a Nix derivation and push the prebuilt .a bundle directly to Cachix    
  (see .github/workflows/nix-build.yml https://github.com/fedimint/fedimint-sdk-ffi/blob/main/.github/workflows/nix-build.yml).                                                                             
                                                                                                                                                                                                            
  The main benefit is that downstream Swift-only PRs can download the precompiled binary from Cachix in seconds without compiling Rust/C code at all, giving us true binary caching and matching our Android
  setup.                                                                                                                                                                                                    
                                                                                                                                                                                                            
  Could you take a look and see if we can adapt that pattern into nix/ffi.nix?

sadeeqrabiu:
> Hey @sadeeqrabiu,
> 
> One quick note on caching: take a look at how we handled the iOS build in fedimint/fedimint-sdk-ffi flake.nix https://github.com/fedimint/fedimint-sdk-ffi/blob/main/flake.nix#L110-L180.
> 
> By using __noChroot = true (relaxed sandbox) and a small xcode-wrapper symlink derivation, we were able to build the iOS slices as a Nix derivation and push the prebuilt .a bundle directly to Cachix (see .github/workflows/nix-build.yml https://github.com/fedimint/fedimint-sdk-ffi/blob/main/.github/workflows/nix-build.yml).
> 
> The main benefit is that downstream Swift-only PRs can download the precompiled binary from Cachix in seconds without compiling Rust/C code at all, giving us true binary caching and matching our Android setup.
> 
> Could you take a look and see if we can adapt that pattern into nix/ffi.nix?

ofc, we can and should adapt that exact pattern into nix/ffi.nix.

MrImmortal09:
@fedimint-bot review again, and check if any of the past review points is missing to be completed.

ios/Demo/DemoModel.swift
```diff
@@ -0,0 +1,648 @@
...
+
+            // Quoting is a round trip, and the field is editable throughout it.
+            // The `didSet` on `lnInvoice` clears any existing quote the moment
+            // the user types, but it cannot stop a call already in flight from
+            // landing afterwards — and arming Pay with a quote for an invoice
+            // that is no longer on screen is how the wrong invoice gets paid.
+            guard invoice == self.lnInvoice.trimmed else {
+                return "the invoice changed while quoting — quote again"
+            }
+            self.lnQuote = quote
```
zeenix: **[P2] Check cancellation before publishing the quote**

Quote twice without changing the input, and let the newer request finish first. The cancelled
older request still passes this input check and assigns `lnQuote`; `run()` only suppresses its
result text after the block returns. Pay therefore uses the older quote while the displayed
fee/route/expiry belongs to the newer one. A failed re-quote can similarly leave Pay re-enabled.
The on-chain quote has the same problem.

Please check cancellation/current-request identity after the await and before publishing the
quote. The same protection is needed for other state writes and watcher installation inside
these blocks.

ios/Demo/DemoModel.swift
```diff
@@ -0,0 +1,648 @@
...
+        // All of this names values from the federation being replaced, and the
+        // copy buttons are gated on it being non-nil — so leaving it in place
+        // offers the user the previous federation's invoice, address and notes
+        // under the new federation's screen. `lastNotes` is the one that
+        // matters: e-cash is a bearer token, and those notes are drawn on a
+        // different mint.
+        lastInvoice = nil
+        lastNotes = nil
+        lastAddress = nil
+        for section in Section.federationScoped { results[section] = nil }
```
zeenix: **[P2] Invalidate pending actions when switching federations**

Clearing these values does not invalidate the `runTasks` already awaiting calls on the previous
federation. For example, start `deposit()` on A, then join/attach B before `onchain.receive()`
returns. A's call subsequently repopulates `lastAddress`, installs its watcher, and renders its
result under B. Pending invoice/ecash requests and quotes can also repopulate the cleared state.

Please invalidate federation-scoped runs on attachment and check that identity before their
post-await mutations. The join/open caller can be preserved separately; leaving every run active
does not prevent cross-federation results.

ios/README.md
```diff
@@ -0,0 +1,288 @@
...
+`binaryTarget` to resolve — see [Publishing](#publishing) for what hosting one
+would take.
+
+```swift
+dependencies: [
+    .package(path: "../fedimint-sdk/ios")
+],
+targets: [
+    .target(name: "MyApp", dependencies: [
+        .product(name: "FedimintSdk", package: "FedimintSdk")
```
zeenix: **[P2] Match the local dependency's SwiftPM identity**

With `.package(path: "../fedimint-sdk/ios")`, SwiftPM derives the dependency identity from the
directory name, so it is `ios`. This product reference instead asks for `FedimintSdk`, causing
an "unknown package" error in the consuming manifest even after the XCFramework is built.
The dependency's own `Package(name:)` does not override that identity.

Use `package: "ios"` here, or explicitly name the local dependency:
`.package(name: "FedimintSdk", path: "../fedimint-sdk/ios")`.

zeenix (changes requested):

Reviewed `6fd05fb`. Three P2 findings below: cancelled requests can replace the displayed quote,
pending actions can restore outputs from a previous federation, and the SwiftPM installation
example references the wrong package identity.

Validation: shell syntax and whitespace checks passed, as did script orchestration checks with
mocked Cargo/Apple tools covering stale slices, Intel-only builds, relative paths, and build
failure.

ios/Demo/DemoModel.swift
sadeeqrabiu: Good catch. Addressed in [`647ef04`](https://github.com/fedimint/fedimint-sdk/commit/647ef0448551b5f6d94daee7a95cfeaf492d24f0).

I introduced `private var isSuperseded: Bool { Task.isCancelled }` on `DemoModel` and added cancellation guards immediately after the `await` returns and before publishing state:
- **Lightning Out (`lnQuoteInvoice`)**:
  ```swift
  let quote = try await lightning.quote(invoice: invoice)
  guard !self.isSuperseded else { return "superseded by a newer quote" }
  guard invoice == self.lnInvoice.trimmed else { ... }
  self.lnQuote = quote

ios/Demo/DemoModel.swift
sadeeqrabiu: Resolved in [`647ef04`](https://github.com/fedimint/fedimint-sdk/commit/647ef0448551b5f6d94daee7a95cfeaf492d24f0).
`attach(_ joined:)` now explicitly cancels and clears all running federation-scoped tasks:
```swift
for section in Section.federationScoped {
    runTasks[section]?.cancel()
    runTasks[section] = nil
    results[section] = nil
}

File: ios/README.md (hunk 5 above)
Comment (sadeeqrabiu) [reply ios/README.md]: Resolved in [`647ef04`](https://github.com/fedimint/fedimint-sdk/commit/647ef0448551b5f6d94daee7a95cfeaf492d24f0).
`attach(_ joined:)` now explicitly cancels and clears all running federation-scoped tasks:
```swift
for section in Section.federationScoped {
    runTasks[section]?.cancel()
    runTasks[section] = nil
    results[section] = nil
}

File: ios/Demo/DemoModel.swift (hunk 4 above)
Comment (zeenix) [reply ios/Demo/DemoModel.swift]: The deposit/quote paths are fixed in `6116316`, but one federation-dependent action is still
missed: `refreshBalance()` uses `run(.wallet)` (lines 157–164), and `.wallet` is excluded from
`Section.federationScoped`. Start Refresh on A, attach B, then let A's balance call finish:
the new `isSuperseded` check is false because this task was never cancelled, so A's amount
overwrites B's balance and can remain there until another balance update.

Please give refresh a federation-scoped task identity, or check a captured federation generation
before writing its result, while keeping the wallet-open caller separate.

File: ios/Demo/DemoModel.swift [hunk 6]
```diff
@@ -0,0 +1,701 @@
...
+
+            let quote = try await ecash.quote(amount: msats)
+            let sent = try await ecash.send(quote: quote)
+            // Guarded like the rest so Copy notes always matches the notes the
+            // screen is showing. Worth being clear about what this does *not*
+            // fix, though: two taps here are two real sends, and the superseded
+            // one has already moved value whose notes then appear nowhere in the
+            // UI. The actual fix is to disable the button while a send is in
+            // flight; this guard only keeps the two displays consistent.
+            guard !self.isSuperseded else { return "superseded by a newer send" }
```
zeenix: **[P2] Prevent overlapping ecash sends before discarding their results**

The new guard drops `sent.notes` after the send has already moved value. The Quote + Send button
is still enabled while a request is pending, so a second tap cancels the first run. If the first
send has committed but has not returned yet, its notes and operation ID are then suppressed,
even if the second request fails. The current demo provides no way to retrieve those notes or
request their cancellation from Activity.

Please implement the in-flight guard mentioned here so another tap cannot supersede an active
send. Also check cancellation after `ecash.quote()` and before starting `send()`: a federation
switch during the quote currently still starts a spend on the old federation.

ios/README.md
```diff
@@ -0,0 +1,305 @@
...
+two-step split exists to prevent. Run a full build when you want a full
+framework back.
+
+Non-Nix escape hatch: the scripts are plain `cargo` and never call `nix`, so
+running them directly works on any machine whose toolchain already has the Apple
+targets — the counterpart of `build-android-sdk.sh --local`:
+
+```sh
+rustup target add aarch64-apple-ios aarch64-apple-ios-sim x86_64-apple-ios
+./scripts/build-ios-sdk.sh
```
zeenix: **[P3] Pass --local in the non-Nix example**

The default now invokes `nix-build-ios-lib.sh`, so this command fails with
`nix: command not found` on the machine this section describes. Please change it to
`./scripts/build-ios-sdk.sh --local` and update the following paragraph, which still says CI
does not use Nix.

.github/workflows/swift-sdk.yaml
```diff
@@ -0,0 +1,157 @@
...
+
+  pull_request:
+    paths:
+      - 'ios/**'
+      - 'rust/fedimint-sdk/**'
+      - 'rust/uniffi-bindgen/**'
+      - 'scripts/build-ios-lib.sh'
+      - 'scripts/nix-build-ios-lib.sh'
+      - 'scripts/generate-swift-bindings.sh'
+      - 'scripts/build-ios-sdk.sh'
```
zeenix: **[P3] Trigger Swift CI when the extracted assembly script changes**

`scripts/assemble-ios-xcframework.sh` now owns XCFramework assembly, but it is missing from
both the pull-request and push path filters. A PR changing only that script therefore runs
neither the assembly check nor the Swift tests/demo build. Please include the new path in both
lists.

zeenix (changes requested):

Rechecked `6116316`. The quote-publication guards and SwiftPM dependency example are fixed.
The Nix/Cachix pipeline is now implemented; all checks pass, including the four native targets,
10 Swift tests, and the simulator demo build.

Two P2 demo issues remain: Refresh still survives federation switches (explained in the reopened
thread), and a superseded ecash send discards its issued notes from the UI. Two smaller build/docs
gaps are noted inline.

Shell checks and mocked producer/assembly checks passed for Nix/Cargo switching, subset builds,
and failure cleanup. The documented non-Nix command was confirmed to require `--local`.

.github/workflows/swift-sdk.yaml
sadeeqrabiu: Added `- 'scripts/assemble-ios-xcframework.sh'` to both `pull_request.paths` and `push.paths` filters in swift-sdk.yaml.

ios/Demo/DemoModel.swift
zeenix: The synchronous in-flight guard and pre-send cancellation check fix double taps and switching
during quoting in `e006ba7`. One cancellation path remains in `ecashSend()` (lines 326–345):

Start a send on A, attach B while `ecash.send()` is awaiting, then let the send succeed.
`attach()` cancels that run and clears its result. The new write restores A's `lastNotes`, but
`run()` still discards the returned success text at line 618 because the task is cancelled.
Copy notes now offers A's notes under B without the amount, operation ID, or any explanation
that this is A's completed send. The operation text promised by the comment never appears.
If the switch happens during `sent.operation.state()` instead, `attach()` clears the newly stored
notes too, and the cancelled run never restores either the notes or the result.

Please preserve the committed send's notes and result together, labelled with their originating
federation, independently of the cancelled run; alternatively, prevent attachment while the
send is committing. The unguarded `lastNotes` assignment alone does not cover these paths.

zeenix (changes requested):
Rechecked `e006ba7308b1506f45838843d26a0133157f680f`.

Confirmed fixed: refresh after a federation switch, overlapping ecash sends, cancellation
between quote and send, the non-Nix build example, and both assembly-script CI path filters.

One P2 remains in the reopened ecash thread: a federation switch during a committed send still
suppresses its result and operation ID, and can clear its notes during the final state lookup.
Please preserve the complete result or prevent that switch while the send is in flight.

All four CI workflows pass at this head. Swift CI reports 10 tests with zero failures and a
successful simulator demo build. Those package tests do not exercise the demo's UI interleavings.

PR #402, opened 2026-09-21
feat(sdk): compile the uniffi surface for wasm32

zeenix:
## Description

- `rust/fedimint-sdk`'s `uniffi` feature now compiles for `wasm32-unknown-unknown`. Before, turning it on with the wasm target died in tokio's feature check, because the feature pulled in `uniffi/tokio` on every target
- `uniffi` is declared once per target: native keeps `tokio` (plus `async-compat`'s multi-thread fallback), wasm gets `wasm-unstable-single-threaded`, which drops the `Send + Sync` bounds the wasm client types cannot meet, and the `uniffi-runtime-wasm` helper crate the browser player calls for its allocator and panic hook
- every exported `async fn` carries `async_runtime = "tokio"` on native only. On wasm the JavaScript host drives UniFFI's future ABI from its own event loop
- `create_fedimint_sdk`'s `data_dir` is a directory path on native and the name of an origin-scoped browser store on wasm
- CI gains `cargo check --target wasm32-unknown-unknown --features uniffi`
- the browser package generated from this surface follows in a separate PR

## Testing

- release wasm32 build with the feature: 23.8 MB module
- unit, integration and doc test counts unchanged


zeenix (commented):
Reviewed by Codex on behalf of @zeenix. One P1 regression, detailed inline.

rust/fedimint-sdk/Cargo.toml
```diff
@@ -19,19 +19,16 @@ publish = false   # flip when the name is reserved and the API settles
...
+# The UniFFI surface: `#[uniffi::export]` on the real methods plus the adapters in
+# `src/ffi/`. Off by default so plain-Rust users compile none of it. The `uniffi` crate is
+# declared once per target below, because the two hosts drive an exported `async fn`
+# differently: a mobile host polls it from a foreign thread, so native needs `uniffi/tokio`
+# (an `async-compat` context around every poll; see that entry) and the exports say
+# `async_runtime = "tokio"`; a JavaScript host drives UniFFI's own future ABI from its event
+# loop on a single thread, so wasm needs `wasm-unstable-single-threaded` (which drops the
+# `Send + Sync` bounds the wasm client types cannot meet) and the exports carry no runtime.
+# `uniffi-runtime-wasm` is the allocator and panic hook the wasm player calls into.
+uniffi = ["dep:uniffi", "dep:uniffi-runtime-wasm"]
```
zeenix: **P1: Keep the native multithreaded fallback enabled.**

`uniffi` no longer activates the optional `async-compat` dependency, so its `multi-thread`
feature stays off with `--features uniffi`. Native bindings fall back to a current-thread
runtime; joining or reopening a federation spawns database work there, where RocksDB's
`block_in_place` panics and aborts release mobile builds.

Add `"dep:async-compat"` here; its target-specific declaration keeps it off wasm. Confirmed
with the SDK's dependency tree and a minimal UniFFI Compat reproduction: it panics without
activation and passes with it.

zeenix: Fixed: the feature now lists `dep:async-compat`, and `cargo tree -e features -i async-compat --features uniffi` shows `multi-thread` enabled again. The feature comment says why the activation matters.

zeenix (commented):
Rechecked by Codex on behalf of @zeenix at `70b9308`.

The native runtime finding is fixed: `uniffi` activates `async-compat/multi-thread`, the
Compat runtime reproduction passes, and wasm still excludes that dependency. No new findings.

zeenix:
@fedimint-bot review

MrImmortal09:
Our old wasm is 10.5MB, significant increase over it.
Could we optimize it further?

zeenix:
> Our old wasm is 10.5MB, significant increase over it.
> Could we optimize it further? 

I'm confused. The report says: 

> Total Size: 8.55 MB

zeenix:
Both numbers are right: 10.5 MB is the raw file, 5.68 MB in the size report is gzipped. Measured against the previous client:

| module | raw | gzipped |
| --- | --- | --- |
| previous `fedimint_client_wasm_bg.wasm` | 11.5 MB | 4.3 MB |
| `fedimint_sdk.wasm` as shipped until now | 21.1 MB | 5.7 MB |
| `fedimint_sdk.wasm` after `wasm-opt -Oz` | 12.3 MB | 4.5 MB |

A third of the shipped module was the `name` section (debug function names). wasm-pack ran `wasm-opt -Oz` on the previous client; this pipeline stopped after wasm-bindgen. #403 now runs `wasm-opt -Oz` as the last step of the generation script (a38bb83), and the browser suite passes on the result. The remaining 7% is real code: the uniffi runtime, the SDK's own facades and the v2 module clients the previous crate never linked.

On build time, measured on a 96-core machine:

- dev profile (`cargo build`, `cargo check`, `cargo test`): full 35 to 68 s, after editing one source file 1 to 5 s. Nothing heavy is enabled there.
- release profile: the crate uses fat LTO with one codegen unit, so a release rebuild after an edit takes 25 s here and longer on fewer cores; the previous client relinked in 2 s because it was a thin shim over the client crates with thin LTO.
- `just generate-sdk-web-bindings` reads the nix-built release module. Both it and its dependency derivation are in fedimint.cachix.org for this branch (checked the narinfos), so with that cache configured nothing compiles; a local build happens only when Cargo.toml or the lockfile change, or when nix is not trusting the cache.

zeenix (commented):
Rechecked by Codex on behalf of @zeenix at `90cc9ec`.

The rebased patch has the same stable patch ID as the previously reviewed `70b9308`; the native
runtime fix and wasm feature gating are unchanged. No new findings.

zeenix:
@MrImmortal09 Let's merge this one, shall we?

MrImmortal09 (approved):
LGTM

PR #403, opened 2026-09-21
feat(sdk): the browser and React Native bindings over rust/fedimint-sdk (T14)

zeenix:
## Description

- stacked on https://github.com/fedimint/fedimint-sdk/pull/402
- `@fedimint/sdk-web` (`js/web/sdk-web`, private for now): the SDK crate compiled to WebAssembly and driven from TypeScript that `uniffi-bindgen-react-native` generates from the crate's `#[uniffi::export]` annotations. Nothing hand-written mirrors the API
- the generator comes from the fork branch that reads UniFFI 0.32 metadata (jhugman/uniffi-bindgen-react-native#468), packaged by nix together with `wasm-bindgen` 0.2.106, the version the crate's dependency tree links. Its `wasm2` flavor reads the metadata out of the `.wasm` itself, so there is one cargo build and no generated shim crate. The runtime packages (`@ubjs/wasm`, `@ubjs/core`, `uniffi-runtime-wasm` 0.31.0-5) are unchanged by the fork and used as published
- the module is built as the cached nix derivation `.#fedimint-sdk-wasm`, the same split the Android SDK uses: nix builds the library, a script reads the bindings out of it
- the SDK runs in a dedicated Web Worker because its browser storage is an origin-private file opened through a synchronous access handle, which only workers get. The page talks to it through a small handle protocol: SDK objects stay in the worker and cross as handles, records and enums cross by value, nested objects are found by a deep walk. Every method is therefore `await`ed on the page, handles die with `close()`, and tagged enums arrive as `{ tag, ...fields }`. When the worker dies, every call in flight rejects
- the generated TypeScript is committed so the JS-only CI jobs work without nix and API changes show up in review; the `.wasm` is a build artifact. The web test job rebuilds, regenerates and fails when the three ubrn-written files are stale. The wasm-bindgen glue is not compared: that tool orders closure exports differently on every run
- a change under `rust/fedimint-sdk` now also runs the web test job, since the bindings derive from it
- `nix/ffi.nix` passes `extraRustFlags = ""` explicitly because flakebox's wasm32 target has no default for it
- the React Native bindings package is generated from `rust/fedimint-sdk` by the same CLI, over the nix-built Android libraries (`.#fedimint-sdk-android-jni`, `--no-cargo`); its generated TypeScript and C++ are committed and checked for freshness like the web ones. The npm `uniffi-bindgen-react-native` dependency moves to 0.31.0-5 for the C++/JSI runtime headers only; its own CLI is not used, since it generates for UniFFI 0.31
- `@fedimint/react-native` opens the SDK through those bindings (`openSdk`, same shape as the web package minus the worker); the JSON-RPC transport over `fedimint-client-uniffi` goes with it
- the two React Native example apps (bare and Expo) are ported to that API: seed, join, balance updates, ecash redeem, lightning quote and pay, invoice receive, on-chain receive and send, activity; the RN unit job type-checks them against the built packages
- the three RN CI lanes #388 deleted are back: unit tests and typecheck against a stub, the Android build on the self-hosted runner, the iOS build on `macos-latest` behind the `ci-ios` label. The iOS lane is restored from the pre-#388 recipe and has not run yet

## Not in this PR

- `wasm-opt`; the release module is about 24 MB
- a per-handle proxy cache on the page: the generated API constructs a fresh wrapper per return, so two proxies never share a handle today

## Testing

- browser test against devimint (`sdk-web` vitest project, part of `pnpm test`): join, lightning receive claimed, balance, restart shows the federation with its balance, closed session rejects and closes idempotently
- unit tests for the protocol, the session lifecycle (worker death fails every pending call) and the type mapper
- `openSdk` of the React Native package against a stub of the bindings; the Android generation run twice is byte-stable


js/web/sdk-web/scripts/copy-wasm.mjs
```diff
@@ -0,0 +1,15 @@
...
+
+const src = new URL('../src/generated/fedimint_sdk.wasm', import.meta.url)
+const dst = new URL('../dist/generated/fedimint_sdk.wasm', import.meta.url)
+if (!existsSync(src)) {
+  console.warn(
+    'sdk-web: no src/generated/fedimint_sdk.wasm; run pnpm generate (needs nix)',
+  )
+} else {
+  mkdirSync(new URL('../dist/generated/', import.meta.url), { recursive: true })
+  copyFileSync(src, dst)
```
zeenix: [P1] Copy the wasm-bindgen JavaScript glue into dist

The build emits `dist/generated/index.js`, which imports `./fedimint_sdk_bg.js`, but `tsc` has no `allowJs` and this script only copies the `.wasm`. Consequently the built package has no `dist/generated/fedimint_sdk_bg.js`: bundling the generated entry fails with `Could not resolve './fedimint_sdk_bg.js'`, so the exported worker cannot load even when the wasm artifact is present. Copy the glue into `dist/generated` as well and exercise the built package entry; the current browser test imports the source tree and misses this.

js/examples/expo-app/app/(tabs)/settings.tsx
```diff
@@ -12,90 +19,69 @@ import {
...
-      const mnemonic = await director.getMnemonic()
-      if (mnemonic && mnemonic.length > 0) {
-        setMnemonicState(mnemonic.join(' '))
-        setMessage({ text: 'Mnemonic retrieved!', type: 'success' })
-        setShowMnemonic(true)
-      } else {
-        setMessage({ text: 'No mnemonic found', type: 'error' })
-      }
+      await reset()
+      await open(restoreWords)
```
zeenix, line 61: [P1] Validate the replacement mnemonic before deleting wallet data

`reset()` shuts down the current session and deletes its entire data directory, while mnemonic validation only happens afterwards in `openSdk()` via `Mnemonic.fromWords()`. Any nonempty invalid phrase, including a checksum typo, therefore erases the existing seed and wallet state before reporting an error, without establishing a replacement wallet. Validate the entered phrase before calling `reset()` and leave the current wallet intact on validation failure. The same ordering exists in `js/examples/react-native/src/App.tsx`.

js/examples/expo-app/app/send.tsx
```diff
@@ -11,19 +18,50 @@ import {
...
-    } catch (e) {
-      setLightningError(String(e))
-      setLightningResult('')
+      const fed = federation()
+      if (!fed) throw new Error('Join a federation first')
+      const lightning = fed.lightning()
+      if (!lightning) {
+        throw new Error('Lightning is not supported by this federation')
+      }
+      setQuote(await lightning.quote(invoice.trim()))
```
zeenix: [P1] Discard quotes whose inputs changed while the request was pending

The input remains editable while `lightning.quote()` is pending. Enter invoice A, start Quote, change the field to invoice B, then let A's request resolve: `onChangeText` clears the quote, but this continuation reinstalls A's quote. The screen shows B while Pay calls `send()` with A's quote, potentially paying the wrong recipient. I reproduced that sequence in both examples. Associate each response with the current input/request generation, or prevent edits until the request settles. The on-chain amount/address quote flow and the bare React Native equivalents have the same race.

js/web/sdk-web/src/types.ts
```diff
@@ -0,0 +1,58 @@
...
+    : V extends Array<infer E>
+      ? Array<ProxiedValue<E>>
+      : V extends Map<infer K, infer W>
+        ? Map<ProxiedValue<K>, ProxiedValue<W>>
+        : V extends { readonly tag: string }
+          ? Fields<V>
+          : V extends object
+            ? HasMethod<V> extends never
+              ? { [K in keyof V]: ProxiedValue<V[K]> }
+              : Proxied<V>
```
zeenix: [P2] Preserve AbortSignal in proxied async options

Generated async methods accept a trailing `{ signal: AbortSignal }`, but this structural mapping treats `AbortSignal` as an SDK object because it has methods. The public signature consequently expects `Proxied<AbortSignal>`: `await updates.next({ signal: new AbortController().signal })` fails with TS2322 because `throwIfAborted()` returns `void`, not `Promise<void>`. Preserve the local AbortSignal/async-options type instead of proxying it so TypeScript callers can use the cancellation path implemented by `splitSignal()`.

js/web/sdk-web/src/client.ts
```diff
@@ -0,0 +1,157 @@
...
+    this.fail(new SessionClosed())
+  }
+
+  private request(head: CallHead, rawArgs: unknown[]): Promise<unknown> {
+    if (this.dead) return Promise.reject(this.dead)
+    const id = this.nextId++
+    const { args, signal } = splitSignal(rawArgs)
+    return new Promise((resolve, reject) => {
+      this.pending.set(id, { resolve, reject })
+      this.worker.postMessage({
```
zeenix: [P2] Reject already-aborted signals before posting a call

A signal that has already been aborted never emits another `abort` event, but `request()` only installs an event listener and unconditionally posts the call. The worker creates a fresh, non-aborted controller, so a call made after `controller.abort()` still runs; an `updates.next()` can remain pending indefinitely. This also differs from `@ubjs/core`, which rejects pre-aborted calls before creating the Rust future. Check `signal.aborted` and reject with an AbortError before registering/posting the request.

js/web/sdk-web/src/client.ts
```diff
@@ -0,0 +1,157 @@
...
+        ...head,
+        id,
+        args: walk(args, (v) => this.markerFor(v)),
+        abortable: signal !== undefined,
+      } as Request)
+      signal?.addEventListener('abort', () => {
+        if (this.pending.has(id))
+          this.worker.postMessage({ kind: 'abort', id } satisfies Request)
+      })
```
zeenix: [P2] Remove the abort listener when a request settles

Each call adds an anonymous listener to the caller's signal, but neither `onMessage()` nor `fail()` removes it. Reusing one controller for a normal `updates.next({ signal })` loop therefore accumulates one retained closure per completed update, along with references to the session; aborting eventually walks all of them. A single successfully resolved call already leaves its listener installed. Store per-request cleanup and run it on success, error, and session shutdown/failure.

zeenix (commented):

Reviewed this second T14 part against the attached implementation plan and the #402 base. Found six introduced issues: the built browser package is missing required glue; restore can erase a wallet before validating its replacement seed; a stale quote can pay the wrong invoice/address; and browser cancellation has a type error, ignores already-aborted signals, and leaks listeners.

Validation: all 19 existing browser-protocol/type and React Native wrapper unit tests passed. Focused checks reproduced the missing-module bundle failure, the public AbortSignal type error, ignored pre-aborts and retained listeners. Executing the actual example component handlers with mocked SDK boundaries reproduced the restore ordering and stale-invoice payment in both examples. Findings are anchored to ab14a7884e1cb3eb710fc8668a61a964637cbb58.

js/web/sdk-web/scripts/copy-wasm.mjs
zeenix: Fixed in be8e757: the copy script now carries `fedimint_sdk_bg.js` into `dist/generated` alongside the module (failing hard when the committed glue is missing), `prepare` runs the full build, and the browser test imports `@fedimint/sdk-web` through the package's `exports`, so the worker entry, the glue and the module are loaded from `dist`. The suite passes against devimint that way (59/59).

js/examples/expo-app/app/(tabs)/settings.tsx
zeenix: Fixed in 0835cf1: both restore handlers parse the phrase with `Mnemonic.fromWords` before `reset()`, so an invalid phrase reports its error and leaves the current wallet untouched.

js/examples/expo-app/app/send.tsx
zeenix: Fixed in 0835cf1: each quote handler (lightning and on-chain, both apps) tags its request with a generation counter that every edit of the inputs bumps; a response for an older generation is discarded before it can be installed, and the inputs are not editable while a quote is pending.

js/web/sdk-web/src/types.ts
zeenix: Fixed in be8e757: `ProxiedValue` keeps `AbortSignal` as itself (it never crosses to the worker; `splitSignal` strips it), so `{ signal }` options type-check. A type test covers `Proxied<BalanceUpdatesLike>['next']`.

js/web/sdk-web/src/client.ts
zeenix: Fixed in be8e757: `request()` rejects with an `AbortError` `DOMException` before posting when the signal is already aborted, and posts nothing. Unit test added.

js/web/sdk-web/src/client.ts
zeenix: Fixed in be8e757: each pending call stores a cleanup that removes its `abort` listener; it runs when the call settles and for every call in flight when the session fails. Two unit tests spy on the signal's add/remove calls and fail if the listener stays.

zeenix (commented):
Rechecked by Codex on behalf of @zeenix at `0835cf1`.

All six findings are addressed. The built browser entry bundles, cancellation types and
cleanup work, and both examples validate mnemonics before reset and discard stale quotes.
No new findings in the updates.

Validation: 23 unit tests and browser typechecking passed. Focused handler checks covered
invalid/valid restores and stale/current Lightning and on-chain quotes in both examples.
The latest browser/devimint CI run also passed.

zeenix:
@fedimint-bot review

zeenix:
Fixed in c93d254: the worker now creates and registers a call's AbortController synchronously on entry, before awaiting the module's initialisation. Since `handle()` is entered synchronously from the message listener, an `abort` for that id cannot be dispatched before the controller is in the map, so a cancellation sent while the module is still loading is honoured. Browser suite against devimint: 59/59.

zeenix:
@fedimint-bot review.

js/.changeset/config.json
MrImmortal09: Could we simply name is web?

scripts/rn-example.sh
```diff
@@ -0,0 +1,70 @@
...
+
+SERIAL="$(adb get-serialno)"
+echo "==> Target: $SERIAL (Android $(adb shell getprop ro.build.version.release | tr -d '\r'))," \
+  "app $PACKAGE from $DIR"
+
+step() { echo "==> [$(date +%H:%M:%S)] $*"; }
+metro_up() { curl -sf http://localhost:8081/status >/dev/null 2>&1; }
+
+if metro_up; then
+  step "Metro is already serving on port 8081"
```
zeenix: [P2] Check the project root before reusing Metro

After running one example, launching the other reuses the first app's detached Metro on port
8081: `metro_up` accepts any successful `/status` response without checking which project it
serves. The second app then requests its bundle from the wrong project and fails to load.
I reproduced the launcher reporting success for `expo-app` with a Metro status response whose
`X-React-Native-Project-Root` pointed to the bare RN example; it never started Expo's Metro.
Verify the running project's root matches `$DIR`, or stop with instructions to switch servers.

scripts/android-emulator.sh
```diff
@@ -0,0 +1,86 @@
...
+set_ini hw.gpu.enabled yes
+set_ini hw.gpu.mode swiftshader_indirect
+
+echo "==> Starting the emulator (device $AVD)"
+emulator "@$AVD" -no-audio -no-boot-anim "$@" &
+EMULATOR_PID=$!
+trap 'kill "$EMULATOR_PID" 2>/dev/null || true' EXIT INT TERM
+
+echo "==> Waiting for Android to boot"
+adb wait-for-device
```
zeenix: [P2] Bound device discovery and monitor the emulator while waiting

If the emulator exits before connecting to adb, `adb wait-for-device` blocks indefinitely.
Both the four-minute deadline and the `kill -0` check are below this call, so neither can detect
that startup failure. A fixture where the emulator exits immediately remains stuck here and
never enters the boot loop. Include device discovery in the bounded, process-monitored wait
so a failed emulator exits the script with an error.

zeenix (commented):
Reviewed by Codex on behalf of @zeenix at `4d2cdbc`.

The earlier fixes remain intact. Found two P2 issues in the new Android scripts, detailed
inline: reusing Metro from another project and hanging when the emulator exits before adb
connects.

Validation: all 23 existing unit tests plus a focused worker-startup cancellation test passed.
Reproduced both script issues with controlled external commands and a local HTTP status server.

scripts/rn-example.sh
zeenix: Fixed in 85c64db: the script reads the `X-React-Native-Project-Root` header of Metro's status response and reuses a server only when it serves this app's directory; a server for the other example (or anything else on the port) stops it with a message naming what to stop. Checked against a leftover bare-app Metro: reused for react-native, refused for expo-app.

scripts/android-emulator.sh
zeenix: Fixed in 85c64db: `adb wait-for-device` is gone; the bounded loop polls `adb get-state` together with the boot property and still checks the emulator process each round. With a fixture emulator that exits at once, the script ends with "the emulator exited during boot" in about two seconds.

scripts/rn-example.sh
```diff
@@ -0,0 +1,83 @@
...
+  "app $PACKAGE from $DIR"
+
+step() { echo "==> [$(date +%H:%M:%S)] $*"; }
+# Metro names the project it serves in a response header; a server left over from the other
+# example answers on the same port but would hand this app the wrong bundle.
+metro_root() {
+  curl -sfI http://localhost:8081/status 2>/dev/null |
+    awk 'tolower($1) == "x-react-native-project-root:" { print $2 }' | tr -d '\r'
+}
+metro_up() { [[ "$(metro_root)" == "$DIR" ]]; }
```
zeenix: [P2] Decode Metro's project-root header before comparing paths

Expo 55's `createMetroStatusMiddleware` emits this header as `encodeURI(metroConfig.projectRoot)`, whereas `$DIR` is a plain filesystem path. For a checkout under `path with spaces`, the same project's header contains `path%20with%20spaces`, so this comparison rejects its own running Metro as belonging to another project. I reproduced the launcher exiting with that false mismatch; the startup polling uses the same comparison and also cannot recognize such a server. Decode the complete header value, or compare it with an equivalently encoded `$DIR`, before checking project identity.

zeenix (commented):
Rechecked by Codex on behalf of @zeenix at `85c64db`.

Both previous findings are fixed: a Metro serving another project is rejected, and an emulator that exits before adb registration now causes a prompt failure. One new P2 remains in the project-root comparison, detailed inline.

Validation: exercised both examples' matching Metro responses, a mismatched project, an unrecognized server, cold startup, emulator early exit, successful boot, and the bounded timeout (with accelerated sleeps). Reproduced the encoded-path regression with a matching-project status response. All active CI jobs passed; iOS was skipped.

scripts/rn-example.sh
zeenix: Fixed in 1e3bd41: the header value is percent-decoded before the comparison (and read as the whole remainder of the line, not the second word, so a plain path with spaces from React Native's Metro survives too). Checked the function against `path%20with%20spaces`, `path with spaces` and a plain path.

scripts/rn-example.sh
```diff
@@ -0,0 +1,87 @@
...
+step() { echo "==> [$(date +%H:%M:%S)] $*"; }
+# Metro names the project it serves in a response header; a server left over from the other
+# example answers on the same port but would hand this app the wrong bundle. Expo's Metro sends
+# the path URI-encoded (a space as %20) and React Native's sends it plain, so the value is
+# percent-decoded before it is compared with the directory.
+metro_root() {
+  local value
+  value=$(curl -sfI http://localhost:8081/status 2>/dev/null |
+    sed -n 's/^[Xx]-[Rr]eact-[Nn]ative-[Pp]roject-[Rr]oot: *//p' | tr -d '\r')
+  printf '%b' "${value//%/\\x}"
```
zeenix: [P3] Preserve literal percent sequences in React Native's plain path

React Native's `statusPageMiddleware` sends `process.cwd()` verbatim, so decoding every header
also changes legitimate directory names. With a checkout under `review%20`, its own Metro
reports that literal path, but this line turns it into `review ` and the launcher refuses the
correct server. I reproduced this with the updated script; the previous comparison accepted
that path. Preserve the raw value for the bare React Native app and decode the Expo value,
or otherwise account for both header formats without rewriting a matching plain path.

zeenix (commented):
Rechecked by Codex on behalf of @zeenix at `1e3bd41`.

The reported spaces issue is fixed for both Expo's encoded header and React Native's plain
header. The earlier launcher fixes remain intact. One minor P3 edge case remains, noted inline.

Validation: 11 script fixture scenarios passed, including encoded UTF-8, matching and mismatched
projects, cold startup, and emulator failure/boot/timeout handling. A further fixture reproduced
the literal-percent edge case. All active CI jobs passed; iOS was skipped.

scripts/rn-example.sh
zeenix: Fixed in 5a80b63: the raw header is compared first and the decoded form only as a second chance, so a plain path with a literal `%20` matches itself and Expo's encoded path still matches its directory. Fixture-checked: plain `review%20`, encoded space, plain spaces, another project, encoded slash, and no server.

Justfile
```diff
@@ -26,9 +26,47 @@ build-android-aar: build-kotlin
...
+# The browser package's bindings, regenerated from the Nix-built wasm module. Run after any
+# change to rust/fedimint-sdk's UniFFI surface and commit the result; CI checks it is fresh.
+generate-sdk-web-bindings:
+    nix develop --accept-flake-config .#wasm-tests -c pnpm --dir js install
+    nix develop --accept-flake-config .#wasm-tests -c scripts/generate-sdk-web-bindings.sh
+
+# The React Native package's bindings, regenerated from the nix-built Android libraries. Run after
+# any change to rust/fedimint-sdk's UniFFI surface and commit the result; CI checks it is fresh.
+generate-sdk-rn-bindings:
+    nix develop --accept-flake-config .#wasm-tests -c pnpm --dir js install
```
MrImmortal09: Why are we using wasm-tests shell to generate the bindings. This would unnecessarily load devimint, which is useless if someone just wants to run android, please check the .#ios and .#android, Also we should have seperate nix shells, It took an hour for me just to build the android(rn) bindings, as wasm-test shell contains much more things than we require. 

I propose to use the existing android and ios shells for building, and then we could have another one with emulator and one more rn-test for ci testing in future.

MrImmortal09 (changes requested):

Justfile
zeenix: Agreed, and done in f6a4e6f. The React Native recipes (`generate-sdk-rn-bindings`, `build-rn-android`), the generation script and the Android CI action now run in `.#android`, which gains `ubrn` (the one tool it lacked; cargo, pnpm and the SDK were already there). The web generator runs in `.#wasm`, which already had ubrn, wasm-bindgen and wasm-opt. `.#wasm-tests` is used only by the browser tests and the devimint integration tests, the things that need devimint. `.#ios` already had ubrn and is unchanged; `.#android-emulator` stays the Android shell plus the emulator, as you propose. Verified both generators reproduce the committed output byte for byte from their new shells.

.github/workflows/verify.yml
```diff
@@ -117,19 +117,19 @@ jobs:
         with:
           cachix_auth_token: ${{ secrets.CACHIX_AUTH_TOKEN }}
 
+      - name: Install JS dependencies
+        shell: bash
+        run: nix develop --accept-flake-config .#wasm-tests --command pnpm --dir js install
```
MrImmortal09, line 122: Why are we using wasm-test shell here just to install the js dependencies?

The default nix shell has node js and pnpm, we could use that.

zeenix (commented):
Rechecked by Codex on behalf of @zeenix at `f6a4e6f`. No new findings.

The literal-percent path issue is fixed. All 12 launcher fixture checks pass, including plain
and encoded paths, spaces, literal percent sequences, and the earlier launcher regressions.

Also reviewed the platform-shell changes: the Android shell provides the required generator,
and the recipes, CI actions, scripts, and documentation consistently use the intended shells.
Android and web binding generation passed CI. All active CI jobs passed; iOS was skipped.

MrImmortal09 (changes requested):

MrImmortal09:
@fedimint-bot start a fresh review, and also point out incorrect shells uses particularly just like my few last comments.

.github/workflows/verify.yml
MrImmortal09: This particularly takes abt 15 mins if not cached.

zeenix:
Fixed in 78f10d6: the web test job's `pnpm install` step runs in the default shell (pnpm and node), and `.#wasm-tests` is entered only by the step that builds and runs the devimint-backed tests. The remaining `.#wasm-tests` uses are the browser test step, the SDK integration tests and `just test`, all of which run devimint.

zeenix:
@fedimint-bot review again

js/examples/expo-app/app/send.tsx
```diff
@@ -67,49 +177,225 @@ const SendOnchain = () => {
...
+  // A quote belongs to the input it was requested for: this counts each request so a quote
+  // that resolves after the amount has since changed can tell it is stale and drop itself.
+  const quoteGen = useRef(0)
+
+  const handleQuote = async () => {
+    const gen = ++quoteGen.current
+    setQuoting(true)
+    setError('')
+    setNotes('')
+    setStatus('')
```
zeenix: [P2] Keep outstanding ecash accessible when requesting another quote

After a successful send, enter `0` and press Quote. These resets erase the displayed notes
before the SDK rejects the amount; because Copy and Cancel send are inside `!!notes`, both
controls disappear too. The first send still holds debited funds, and neither example offers
a history action to retrieve its notes or cancel it. I reproduced this in both apps. Preserve
outstanding sends and their controls, or require explicit cancellation/dismissal, rather than
clearing them as a side effect of quoting. The same reset is in the bare React Native component.

js/examples/expo-app/app/send.tsx
```diff
@@ -67,49 +177,225 @@ const SendOnchain = () => {
...
+      // the notes are reclaimed: by the deadline in the details, or sooner through Cancel
+      // send, which fails against notes the receiver already redeemed and settles the state
+      // as Redeemed.
+      const details = await handle.operation.details()
+      const reclaimAt = new Date(Number(details.reclaimAt)).toLocaleString()
+      setStatus(`Outcome settles by ${reclaimAt}, or when you press Cancel send`)
+      handle.operation
+        .awaitFinal()
+        .then((state) => setStatus(EcashSendState[state]))
+        .catch((error) => setStatus(errorMessage(error)))
```
zeenix: [P2] Associate asynchronous status updates with their send

This callback remains active after another send replaces `notes` and `operation`. Send A,
request its cancellation, then send B before A settles: when A resolves to Canceled, B's
still-spendable notes are displayed with the status Canceled. A Redeemed result similarly
misreports B as paid. I reproduced this in both apps. Guard status/error continuations with
the operation identity or generation, or keep separate state per operation; the cancellation
continuation also needs that ownership check.

js/examples/expo-app/app/send.tsx
```diff
@@ -67,49 +177,225 @@ const SendOnchain = () => {
...
+      setSending(false)
+    }
+  }
+
+  const handleCancel = async () => {
+    if (!operation) return
+    setError('')
+    try {
+      await operation.requestCancel()
+      setStatus('Cancel requested; unredeemed notes return to this wallet')
```
zeenix: [P3] Preserve a final outcome when cancellation is a no-op

Cancel send stays enabled after awaitFinal has displayed Redeemed or Canceled. The SDK
documents requestCancel on a final operation as a successful no-op, but this continuation
replaces that final result with Cancel requested. Since awaitFinal has already completed,
nothing restores the real result. Reproduced in both apps. Disable cancellation for final
operations and prevent a late cancellation acknowledgement from overwriting a final state.

zeenix (commented):
Rechecked by Codex on behalf of @zeenix at `a0e17d1`.

The rebased series has the same tree as the previously reviewed version before the three new
commits, so the earlier fixes are preserved. The default-shell JS install addresses the shell
review comment.

Found two P2 issues and one P3 in the new ecash-send UI, each present in both example apps:
an unsuccessful quote hides outstanding notes and cancellation controls; an older operation
can overwrite a newer send's status; and cancellation can overwrite an already-final status.

Validation: all 23 existing unit tests passed. React component fixtures using the actual
SendEcash components and controlled SDK promises reproduced all three issues in both apps.
All active CI jobs passed; iOS was skipped.

js/examples/expo-app/app/send.tsx
zeenix: Fixed in dbf350b. Quoting no longer clears anything. Each send is now its own entry in a list (newest first) with its own notes, status and controls, so an outstanding send stays visible and cancellable until it settles or you dismiss it.

js/examples/expo-app/app/send.tsx
zeenix: Fixed in dbf350b. Every send gets an id at creation; the details status line, the awaitFinal continuation and the cancel acknowledgement all update only that id's entry, and the operation handles live in a map keyed by the same id.

js/examples/expo-app/app/send.tsx
zeenix: Fixed in dbf350b. A settled send offers Dismiss instead of Cancel send, and the cancel acknowledgement is applied inside the state updater only while that send is not final, so a late acknowledgement cannot overwrite a final state.

.github/workflows/verify.yml
zeenix: Done earlier in this round: the install step now runs in the default shell (`nix develop --command pnpm --dir js install`), and `.#wasm-tests` is entered only for the devimint-backed test step that needs it.

zeenix (commented):
Rechecked by Codex on behalf of @zeenix at `dbf350b`. No new findings.

All three ecash UI findings are fixed in both examples: failed quotes preserve outstanding
sends, asynchronous updates stay with their own send, and late cancel replies preserve final
outcomes. Settled sends offer Dismiss instead of Cancel send.

All 12 focused React component scenarios passed with controlled SDK promises, including the
previous reproductions, delayed cancellation replies, observation errors, and per-send
copy/cancel/dismiss controls. All active CI jobs passed; iOS was skipped.

MrImmortal09:
@fedimint-bot review

MrImmortal09 (approved):
LGTM

Issue #404, opened 2026-09-21
Federation discovery: fetch a list of joinable federations from an app-supplied endpoint


zeenix:
## Motivation

Wallet apps built on this SDK show users a list of federations they can join. The SDK currently covers everything from the invite code onward (`Sdk::preview`, `Sdk::join`), but nothing before it: every app hand-rolls fetching a remote JSON document, filtering it, and turning entries into invite codes.

The Fedi app, for example, fetches such a list from its own web backend and filters it in TypeScript, entirely outside the Rust layer. Other wallets do the same. This is generic, bug-prone plumbing (HTTP, JSON, invite code parsing, expiry filtering, federation ID checks) that fits the high-level API this crate is converging on.

## Proposal

Add a small discovery API to `fedimint-sdk` that fetches a list of federations from a URL the app supplies, validates the entries, and returns typed values that plug straight into `preview`/`join`.

```rust
pub struct DiscoveredFederation {
    pub id: FederationId,
    pub invite: InviteCode,
    /// Publisher-supplied metadata for the entry (e.g. `federation_name`,
    /// `federation_icon_url`, `welcome_message`, `popup_end_timestamp`).
    /// Same key conventions as `FederationPreview::metadata`, but unverified:
    /// this is what the list publisher claims, not what the guardians serve.
    pub metadata: BTreeMap<String, String>,
}

impl Sdk {
    pub async fn discover_federations(&self, url: &str) -> Result<Vec<DiscoveredFederation>>;
}
```

Design points:

- **No default endpoint, ever.** The URL is always app-supplied. This is a general-purpose open source SDK and must not embed any organization's list.
- **Validation in the SDK:** parse invite codes, check the entry's federation ID against the invite code, drop entries whose `popup_end_timestamp` has passed, and skip malformed entries instead of failing the whole list. Ideally optionally exclude federations already joined (`Sdk::federations()`).
- **Document the JSON schema** in this repo. Reuse upstream's meta field conventions (`docs/meta_fields/` in fedimint/fedimint) so publishers already producing `meta_override_url` documents can reuse them.
- **Errors** map onto the existing `ErrorCode` set: `Timeout`, `InvalidInput` for an unparseable document, plus probably one new code for an unreachable/non-2xx endpoint since `FederationUnreachable` is about guardians.
- **Fetch only.** No caching or persistence in the first iteration; apps have their own state layers. A list of URLs with merge and dedupe by federation ID could follow.
[...]

PR #405, opened 2026-09-22
feat(sdk): federation discovery from an app-supplied endpoint

Yunusabdul38:
Implements `Sdk::discover_federations` per #404. **Draft**: the schema question at the bottom should settle before a proper review, since it decides the parser's shape.

## What's here

```rust
pub struct DiscoveredFederation {
    pub id: FederationId,
    pub invite: InviteCode,
    pub meta: BTreeMap<String, String>,
}

impl Sdk {
    pub async fn discover_federations(&self, url: &str) -> Result<Vec<DiscoveredFederation>>;
}
```

A JSON object keyed by federation id; `invite_code` required, every other key carried into `meta` unchanged. Entries are dropped rather than reported when the invite code doesn't parse, when its federation id disagrees with the key it's filed under, or when `federation_expiry_timestamp` has passed. A malformed entry costs that entry, not the list.

Both answers from the issue are applied: the field is `meta` to match `FederationPreview`, and expiry keys off `federation_expiry_timestamp` only, with a test asserting a vendor key like `fedi:popup_end_timestamp` is carried through untouched rather than acted on.

`ErrorCode::EndpointUnreachable` is new, for an app-supplied endpoint that is unreachable or answers non-2xx. `FederationUnreachable` means guardians, so reusing it would have been misleading.

## Following what the crate already does

* Existing types and their validating parses (`FederationId`, `InviteCode`, `Timestamp`) rather than strings or raw integers in the public API.
* Per-target dependencies in the style of `storage.rs`, so "this differs per platform" keeps one pattern.
* reqwest declared at fedimint's own version and feature set, so cargo unifies it onto the copy `fedimint-client` already links. **The lockfile moves by exactly one line** and no second HTTP or TLS stack enters the graph. The wasm target takes the same crate with no features, where it wraps the browser's `fetch`.
* The seconds-to-milliseconds conversion happens once, where the document is parsed, pinned by a test.
* The caller-facing contract lives on `Sdk::discover_federations` rather than on the module, which is private and would reach no reader.

## Tests

15, all pure: the fetch is a closure seam (`discover_with`), so they serve a fixture rather than standing up a server. They cover the happy path, scalar coercion, the id cross-check, missing and malformed invite codes, one-bad-entry-doesn't-spoil-the-list, expiry in both directions and exactly on the boundary, an unreadable expiry, a vendor expiry key, id ordering, an empty document, a non-document answer, and a failed fetch.

Local gate clean: `fmt`, `clippy --all-targets`, `clippy --features uniffi`, `cargo doc`, `cargo check --target wasm32-unknown-unknown`, `cargo test` (547 + 19 + 3).

## Two questions before review

**1. What shape should the document be?**

fedimint-observer publishes a federation list at `observer.fedimint.org/api/federations` (17 federations live), and it is not this shape:

```rust
// fmo_api_types::FederationSummary, returned as Json<Vec<FederationSummary>>
pub struct FederationSummary {
    pub id: FederationId,
    pub name: Option<String>,
    pub invite: String,            // not `invite_code`
    pub health: FederationHealth,  // Online | Degraded | Offline
    pub deposits: Amount,
    pub nostr_votes: FederationRating,
    pub last_7d_activity: Vec<FederationActivity>,
}
```

I ran this branch's parser against the live document:

```
InvalidInput — this endpoint did not answer with a discovery document:
invalid type: sequence, expected a map
```

Two things worth weighing:

* The observer's README says the `/federations` API "isn't stable at this point", so tracking its shape carries an ongoing cost.
* Neither `/api/federations` nor `/api/federations/<id>` carries a description, welcome message or icon, so even parsed perfectly it makes a thin browse screen.

Three ways forward:

1. **Accept both shapes here.** Detect array versus object at the top of the parse, accept either `invite`/`invite_code`, and carry the remaining scalar fields into `meta`. A small branch, since the parsing layer is already pure and lenient, plus tests for the second shape. Worth being precise about what survives: `name`, `health` and `deposits` would land in `meta` as strings, while `nostr_votes` and `last_7d_activity` would be dropped, since `meta` is a map of strings and there's no lossless spelling for an object or an array.
2. **Keep this schema strict** and leave observer support to a follow-up.
3. **Keep it strict and have the observer serve an endpoint in this shape.** The documented, stable format then becomes the one publishers target, rather than the SDK tracking an API its own README calls unstable.

Happy with any of them; flagging it because the schema is what other people publish against, and it's much cheaper to settle now than to widen later.

**2. Should `EndpointUnreachable` carry structured detail?**

Right now the HTTP status is only in the message, which the crate says must never be parsed. `ErrorDetails` exists for exactly this (the numbers a caller would otherwise scrape out of prose), and a caller might reasonably retry a 503 but not a 404. I haven't added a variant, since that touches the envelope codec and the generated bindings, so it seemed worth asking rather than assuming.


zeenix:
@Yunusabdul38 As I wrote to you on Discord, there is absolutely no point in you doing this completely with AI. I can do that myself using Fable. Please start from scratch and do it manually yourself. I will be able to tell if you didn't put in the effort.

